PR #2, opened 2020-07-06
Updated backsword  in Low Tech Equipment.eqp

Exxar:
Added skill defaults for Swing and Thrust usages and elaborated upon the damage bonus note so that it only references the Hilt Punch attack.

### DIFF
--- a/Library/Low Tech/Low Tech Equipment.eqp
+++ b/Library/Low Tech/Low Tech Equipment.eqp
@@ -426,7 +426,7 @@
 				<modifier>0</modifier>
 			</default>
 		</melee_weapon>
-		<notes>The Hilt Punch attack receives damage bonuses for whichever of Boxing (p. B182), Brawling (p. B182), or Karate (p. B203) is used to deliver it</notes>
+		<notes>The Hilt Punch attack receives damage bonuses for Brawling (p. B182)</notes>
 		<categories>
 			<category>Melee Weapon</category>
 		</categories>
--- a/Library/Low Tech/Low Tech Equipment.eqp
+++ b/Library/Low Tech/Low Tech Equipment.eqp
@@ -339,6 +339,45 @@
 			<reach>1</reach>
 			<parry>0</parry>
 			<block>No</block>
+			<default>
+				<type>DX</type>
+				<modifier>-5</modifier>
+			</default>
+			<default>
+				<type>Skill</type>
+				<name>Broadsword</name>
+				<modifier>0</modifier>
+			</default>
+			<default>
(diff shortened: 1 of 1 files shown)


Library/Low Tech/Low Tech Equipment.eqp
```diff
@@ -347,8 +386,47 @@
...
+				<modifier>-4</modifier>
+			</default>
+			<default>
+				<type>Skill</type>
+				<name>Sword!</name>
+				<modifier>0</modifier>
+			</default>
 		</melee_weapon>
-		<notes>Attack receives damage bonuses for whichever of Boxing (p. B182), Brawling (p. B182), or Karate (p. B203) is used to deliver it</notes>
+		<notes>The Hilt Punch attack receives damage bonuses for whichever of Boxing (p. B182), Brawling (p. B182), or Karate (p. B203) is used to deliver it</notes>
```
richardwilkes, line 429: The entry for the hilt punch does not seem to have defaults for anything but brawling. Shouldn't it have entries for boxing and karate as well, if this statement is true?

Exxar: Looking at its entry in LT, it seems the Hilt Punch is indeed only usable by DX or Brawling. So it is mentions of Boxing and Karate that should actually be removed.

I actually only adjusted this note because it is displayed under all attack modes, so I found it misleading having it under Swing and Thrust without that qualifier I added.

Exxar: Ehh. So I based my statement of "it seems the Hilt Punch is indeed only usable by DX or Brawling" by the Backsword Hilt Punch only being listed in the table for "BRAWLING or DX", and not in the table for "BOXING, BRAWLING, KARATE, or DX". But following the [5] note in the table, it seems that this note, as it was before my adjustment, is a verbatim copy of that [5] note from LT.

How do you wish to proceed here? I would adjust the note to only mention DX and Brawling, because that seems the intention here. I assume the [5] note mentiones the other skills simply because that same [5] note is also used for other entries, some of which are in the table for "BOXING, BRAWLING, KARATE, or DX".

richardwilkes: I think removing the references to things it doesn't default to is what should happen, since they are non-sensical in this particular context.

Exxar: Then we agree. Fixed.

richardwilkes:
While this has been merged, it won't show up via the update mechanism until I do another "release" of the data.

PR #6, opened 2020-08-14
Re-organise UT Meta traits into 2 files

Legendsmith:
The confusingly named file is just sitting in a folder by itself, and is a template file rather than a library file. It is the only thing inside its folder. better to split it into two files that have clear names inside that folder.
I like this solution as it clearly shows what traits area inside what file and allows for easy use.
*This is the alternative solution to #7, don't merge them both.*


Legendsmith:
I have closed this for the time being seeing as I think that #7 is the superior option now.

PR #7, opened 2020-08-14
Moved meta traits into an advantage library and into base UT folder.

Legendsmith:
No need to have it in its own folder for no reason and make searching harder. There's also no reason for it to be a template file.
This is the alternative to #6 Merge one and close the other.


richardwilkes:
Why the CYBERNETICS AND UPLOADING container moved from the top level? In your other PR, it was split out into its own file and in the current master library version it was also a top-level container. I haven't used anything from UltraTech since a 3rd Edition game I ran in the late 90's, so am not terribly familiar with any of this and don't have a good feel for it one way or the other.

Also, why is nearly everything closed up? Seems like it would be more convenient in the library to have most, if not all, entries fully disclosed for quick browsing-by-scrolling.

Anyway, of the two submissions (#6 and #7), I personally prefer this one (i.e. a single file), but don't have a strong preference so will defer to you as to which style you'd like to see make it into the master library. Pick one and let me know.

Legendsmith:
I have corrected the error with the nesting, it should have been top level. I have also opened the containers except for the traits themselves. I will consult with a few more people.

PR #8, opened 2020-08-15
Alternative way to organize Low-Tech Armor

angelforest:
Instead of having a long-long list like the one in Instant Armor, this library uses equipment modifiers to greatly reduce the number of visible items. The user simply picks the material, then selects the locations.

### DIFF
--- a/Library/Low Tech/Low Tech Armor.eqp
+++ b/Library/Low Tech/Low Tech Armor.eqp
@@ -1,0 +1,17116 @@
+{
+	"type": "equipment_list",
+	"version": 1,
+	"rows": [
+		{
+			"type": "equipment",
+			"version": 1,
+			"quantity": 1,
+			"description": "Cane",
+			"tech_level": "0",
+			"modifiers": [
+				{
+					"type": "eqp_modifier",
+					"version": 1,
+					"disabled": true,
+					"name": "Head",
+					"cost_type": "to_original_cost",
+					"cost": "+10.5",
+					"weight_type": "to_original_weight",
+					"weight": "+3.6 lb",
+					"features": [
+						{
+							"type": "dr_bonus",
+							"amount": 1,
+							"location": "skull"
+						},
+						{
+							"type": "dr_bonus",
+							"amount": 1,
+							"location": "face"
+						}
+					]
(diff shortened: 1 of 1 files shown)


richardwilkes:
I generally like this approach... however, it has the downside of not being clear what the armor is (the material is clear, but you can't quickly determine that "this my helm"). I'm also not understanding (likely because I've yet to actually use the Low Tech armor for anything) why the face & neck protection are singled out to be special, unlike all the rest.

There's also the issue that GCS doesn't allow you to apply DR to many of these extra locations or partial coverage (front, back, left, right, etc.). I'd love to find a good way to address that so that something like this could actually make use of it... right now, far too many entries have no DR listed. For example, the first thing I applied to a new sheet to see what this did was "Layered Leather, Medium" and I selected "Abdomen" as the modifier. This gave me a note about the hit location, but did not alter my DR display at all and also did not mention the DR that was gained to the Abdomen. That doesn't seem useful to me.

One last note: The page references are missing.

angelforest:
>it has the downside of not being clear what the armor is (the material is clear, but you can't quickly determine that "this my helm")
I figured the users can rename equipment if they wanted to (e.g. hauberk would fine mail with torso, shoulders and thighs selected). The alternative would be to have a separate item entry for each location with just one modifier (material) selected. I guess that would be a better choice for quick reference, at least until GCS can display all those extra locations.
I can remake it either way if desired, the entire thing is automatically generated.

>why the face & neck protection are singled out to be special
This was done to avoid clutter within armor entries, and because they are both listed in separate sections in the books. If I were to group equipment by location, those could be added as modifiers to face and neck respectively, but with one caveat - because currently GCS sorts the modifiers in alphabetical order, those modifiers will be mixed with materials.

For now, I'll note DR value for each material and add page references. I'll also add alternative version with armor grouped by location to see how it works in action.

angelforest:
Okay, so I made a variant grouped by location.
I am not particularly happy about how I had to do face/neck protection because e.g. "+20 lb to original weight" and "-50% to original weight" don't interact with each other the way I expected (it seems +X% applies before +X). Regardless, the hacky way works and produces correct values.

PR #13, opened 2020-10-04
Added Magic: Death Spells and Norðlondr Fólk

rinickolous:
Added the full contents of **Magic: Death Spells** and **Norðlondr Fólk**.

There are currently no page references for **Norðlondr Fólk**, but **Magic: Death Spells** uses **MDS**, which is what I assume is appropriate based on **Magic: Plant Spells** being abbreviated as **MPS**.

This fork also contained changes to some Output Templates (For Fantasy Grounds: Adding Inventory data to NPCs, and putting currency in the main inventory instead of a separate one, to allow for currency with weight). These changes have been reversed and persist only on my private branch.


Library/release.txt
```diff
@@ -0,0 +1 @@
+2.2
```
richardwilkes, line 1: This file should not be checked in. It is for the deployed library, not the git repo.

.gitignore
```diff
@@ -1 +1,4 @@
 /.idea/
+Library/Output Templates/Fantasy Grounds NPC - Inventory.xml
```
richardwilkes, line 2: These changes should be removed

richardwilkes (changes requested):
Generally looks good... but two changes need to be made before I can accept it.

richardwilkes:
I suspect the issue here is that you're editing your installed master library, rather than a separately checked-out git repo of it. What I tend to do is checkout a copy of the master library in my User Library folder, under something uncreatively-named like "gcs_master_library". That way I don't introduce extra bits that aren't supposed to be there and can still edit it from within the program.

Library/release.txt
rinickolous: I think I may have copied the Master Library folder at some point instead of pulling from the repo. Should be fixed now.

.gitignore
rinickolous: Done.

rinickolous:
All request changes should be applied now.

richardwilkes (approved):

richardwilkes:
I don't think I have much, if any, other library changes pending right now. If you have no other changes you plan to make in the next week or so, let me know and I'll tag this so that others will see it.

PR #17, opened 2020-10-21
Minor corrections to races

rinickolous:
- Changed Advantage Container type to "Races" wherever appropriate.
- Renamed some Racial Template files to maintain consistency in naming format (names refer to a single member of the Race, i.e. "Dark Elf" as opposed to "Dark Elves").
- Removed the "Lizard Man" Racial Template from Banestorm (duplicate of "Reptile Man").
- Fixed the price of the Bastard Sword in DFRPG (was $650, now $750).


richardwilkes (approved):
Another update that would be useful is to convert all the notes about reaction modifiers to actual reaction modifiers, since they are now available in GCS

PR #24, opened 2020-11-15
Minor fixes

Legendsmith:
Add weapon damage bonuses from equipment modifiers, to Basic Set and LT libraries.
Add Skill bonuses to Equipment Modifiers for Low Tech using the recent features (thanks Rich!)
Incorrect MA TG technique data for forced posture change, including a duplicate technique.


richardwilkes:
I think this change might need to be broken out into a sub-section for the specific damage types, as the damage bonuses aren't applicable to all weapons. Once broken out, the note should probably be altered to remove the mention of a bonus to damage, as that might confuse players and have them manually add the bonus in each time they strike, when it is already present.

Legendsmith:
Updated tooltips.

richardwilkes (approved):

PR #26, opened 2020-11-18
Fixed Fantasy Ground Output Templates

rinickolous:
The "type" field for spells was outputting an empty string. Corrected to output spell difficulty.
Also changed tabbing to 4-space for consistency with Fantasy Grounds file storage.


richardwilkes:
The fix for #27 addressed the missing key -- your fix came in first, but I failed to notice they were for the same thing, as I started with things attached to the bug reports first.

As for the change from tabs to spaces... I don't know if that's worth it on its own. I'm not against it if you want to fix the conflicts, though.

richardwilkes:
I'm going to close this out, as there has been no response for 3 weeks. Feel free to re-open if you want to move forward.

PR #27, opened 2020-11-18
Fix for "Missing Spell Type Info from FG Export Template"

crnormand:
Fix for https://github.com/richardwilkes/gcs/issues/280


crnormand:
Sorry for all of the extra commits... still very much a GIT newb.

richardwilkes (approved):

PR #28, opened 2020-11-23
Foundry VTT specific XML output template

crnormand:
As of v0.7.0 we can now import either the Fantasy Ground XML export, or this new Foundry export (which contains more data).

Again, I am sorry.   This PR seems to have more commits in it than it should.   I first did a reverse pull request from your repo to bring my github repo up to date.   Then I pulled from github to my PC, added my file, committed and pushed.   I don't understand why it is showing other commits.     If you know a better process, please let me know.


richardwilkes:
You can fix this in the future by either just cloning a fresh copy to work with each time, or I believe you can also just branch off of my main each time to get a similar effect.

richardwilkes:
Do you also want to add support for the new "conditional modifiers" table that gets generated? It is essentially identical to the "reaction modifiers" data.

crnormand:
> Do you also want to add support for the new "conditional modifiers"

If it's OK with you, I'm going to hold off on that. I don't quite have a complete handle on what that means :-)

And I would rather give this to you now, to get it into the pipeline.  ￼￼￼

PR #35, opened 2021-01-16
Minor fixes, add some PU9 modifiers

Legendsmith:
Not sure why it reassigned  a bunch of IDs. Maybe my bad? But yeah here's fixes and some content.


richardwilkes:
Did you duplicate existing entries, then delete the old one? That would cause the seemingly spurious ID changes and additions of the "based_on" attributes.

Legendsmith:
> Did you duplicate existing entries, then delete the old one? That would cause the seemingly spurious ID changes and additions of the "based_on" attributes.

I didn't. I just redownloaded the latest, added the modifier to DR, and then recomitted twice; it still redoes the IDs.

richardwilkes:
Well... that sounds broken. I'll try to make some time soon to take a look and see if this is a real problem or not. Once that's been determined, we can decide whether to keep this as-is or do something to remove those extraneous changes.

richardwilkes (dismissed):

richardwilkes (approved):

PR #40, opened 2021-02-19
Updates for Ritual Magic

trev-dev:
This is a PR that will help update ritual magic spells to adhere to the 2006 publication of GURPS Magic.

Discord conversation for reference: https://discord.com/channels/731926683371700315/731926683371700320/812128414436950036

This particular PR:
1. Removes prerequisites for spells except at least 1 point of Ritual Magic (REF: M200).
2. Updates the individual spells with their up-to-date prerequisite counts. 
3. Adds "Ritual Magery" advantage that is specifically tailored for boosting Ritual Magic and subsequent spell techniques


richardwilkes:
I'm happy for you to break your commits up into as many pieces as you like, but I won't merge the PR until the file itself is complete.

trev-dev:
Sounds good to me @richardwilkes. It's more a mechanism for me to track where I'm at, and for others (assuming anyone else is crazy enough participate) to know where this is at. I will squash the prereq_count commits before signalling you to review :)

trev-dev:
@richardwilkes All updates pushed and description/title of the PR updated. We should be good to go! Do you want an example sheet to help evaluate what's been changed?

richardwilkes:
Example sheet? As in a complete character that uses Ritual Magic? That would be good to add in general for others to see how it works, if you have the time and interest. If you do, you can add it to the "Home Brew/Characters" directory, assuming it isn't a character out of one of the official PDFs, in which case it should go in which directory corresponds to that PDF.

trev-dev:
This wasn't what I meant, I was thinking of sending it to you directly so you could feel confident in the changes, but I like the way you've misunderstood it better.

I would gladly add a "Ritual Magic Example" character file to the repo.

trev-dev:
@richardwilkes Homebrew example added.

richardwilkes:
Should all of the spells be Hard or should some be other difficulty levels, like in the original Magic?

trev-dev:
All of the spells should be Hard Techniques as per M200.

richardwilkes (approved):

PR #42, opened 2021-03-10
V5 of the GCS Export for Foundry

crnormand:
The original export incorrectly assumed there would only be 1 'notes' for a whole weapon.   This update moves the 'notes' export to the modes loop.


richardwilkes (approved):

crnormand:
If it takes more than 5 minutes, then no.   We can wait.   We only have like 13* users :-)   
If it is a simple press of a button and typing a few words of wisdom, then yes, thank you!

PR #46, opened 2021-03-12
Added content from Template Toolkit 2 - Races

kaldrich001:


richardwilkes:
Let's push the "Template Toolkit 1" directory down one level and have a "Template Toolkit" directory at the top, since I know there are several of those out there and presumably we'll eventually have files for them all.

kaldrich001:
A quick skim through TT1 didn't look like anything was there, but I'd forgotten there was a third one. Fixing it now.

richardwilkes (approved):

PR #51, opened 2021-03-16
 Added the Rest in Pieces advantage to the DFRPG advantages library

daliin:


Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Advantages.adq
```diff
@@ -6946,6 +6946,21 @@
 				"Advantage"
 			]
 		},
+		{
+			"type": "advantage",
+			"version": 1,
+			"id": "3717d918-5383-4bb3-a213-7b8b554591cd",
+			"based_on_id": "0eb5e29d-64aa-4a0b-8f9b-87ae7e6997ab",
+			"based_on_hash": "kBs4LfLRDNwXCHIhihMyD2N44WYtx+8Af4+ywt7OGNM",
```
richardwilkes, line 6954: Ideally, neither of these lines would be in the file for this perk. In theory, these are only added if you created the Advantage by duplicating it, then modifying it. Since this is a new perk that isn't based on an existing one, that shouldn't be the case.

You should be able to either use a text editor to remove those two lines manually, of you could open the file up in GCS, delete this perk, then re-create it by creating a new Advantage, then editing it appropriately.

daliin: Ahhh, I see. I actually copied the perk from the Holy Warrior template. I'll delete those lines and resubmit. Thank you!

daliin: I think I've done it properly (I see two commits above, including the two deleted lines). I'm very new to the GitHub process, so let me know if I need to create a new pull request or otherwise click anything.

richardwilkes (approved):

PR #53, opened 2021-03-18
Adding reaction bonus to Social Regard

tiagoaquinofl:


richardwilkes (commented):
This should have the "per level" attribute set to correctly give the bonus

tiagoaquinofl:
I guess now it is..

Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -25189,6 +25189,14 @@
 			"levels": "1",
 			"points_per_level": 5,
 			"reference": "B87",
+			"features": [
+				{
+					"type": "reaction_bonus",
+					"amount": 1,
+					"per_level": true,
+					"situation": "Your class, race, sex (or other group) must be obvious to anyone who meets you; +1 bonus per level, up to +4."
```
richardwilkes, line 25197: Don't really need the explanatory text about "+1 bonus per level" etc, as that's part of the automation

richardwilkes: May also want to put a substitutable marker around the part where it says "class, race, sex (or other group)" so that users are prompted to pick the appropriate thing

tiagoaquinofl:
I guess @Reason@ in notes fulfill that info. Otherwise player would have to input it twice.

richardwilkes:
The note about max levels should be on the advantage, not the feature sub-section.

The "situation" should be made to read logically. For example, it usually starts with something like "from those..." since it would be displayed as "-2 from those that know about your reputation" (as an example).

tiagoaquinofl:
but you know that, because of that I bumped in something I'm opening an issue in the other repository. The variables in the reaction block won't update values until we change something (like upgrade the advantage's level).

Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -25189,6 +25189,14 @@
 			"levels": "1",
 			"points_per_level": 5,
 			"reference": "B87",
+			"features": [
+				{
+					"type": "reaction_bonus",
+					"amount": 1,
+					"per_level": true,
+					"situation": "from those who holds you in high regard for being a @Reason@, in a @Type: Feared, Respected or Venerated@ way."
```
richardwilkes, line 25197: Almost there... “from those who holds you” should be “from those who hold you”.

richardwilkes (approved):

PR #55, opened 2021-03-22
Add "More Skill Sets for Specialists" from Pyramid 112.

kaldrich001:


richardwilkes (approved):

richardwilkes:
In the future, please start with a fresh copy of master before applying your changes -- otherwise, you end up with a long list of old commits here as well, which aren't harmful, but do make it a bit tougher to understand what you changed later when trying to assemble release notes.

PR #58, opened 2021-04-01
Add Illiteracy; adjust Research

daliin:
Added the Illiteracy disadvantage from DFRPG Adventurers (p. 61).
Adjusted the Research skill to replace the prereqs from Basic with the correct one from DFRPG (namely, that the character does not have the Illiteracy disadvantage).

### DIFF
--- a/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Advantages.adq
+++ b/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Advantages.adq
@@ -4760,6 +4760,19 @@
 		{
 			"type": "advantage",
 			"version": 1,
+			"id": "5da5b487-a52b-4418-8e77-09766c232acf",
+			"name": "Illiteracy",
+			"mental": true,
+			"base_points": -3,
+			"reference": "DFA61",
+			"categories": [
+				"Disadvantage",
+				"Language"
+			]
+		},
+		{
+			"type": "advantage",
+			"version": 1,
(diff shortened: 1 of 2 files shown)


richardwilkes:
Hmm... Illiteracy is usually done by specifying the modifier "Written: None" for the given language advantage you've taken. Did they change that in DFRPG?

daliin:
I think they attempted to simplify languages in DFRPG. Per _Adventurers_, p.9, everyone is assumed to speak and read/write Common. If you want to be unable to read/write, then you take the Illiteracy disadvantage for 3 points. Additional languages cost 3 points for spoken or reading/writing and 6 points for both.

richardwilkes:
That's exactly the approach taken in Basic, as far as I'm aware.

I'm not necessarily against this change -- but pointing out it doesn't fit with the current way of handling this. A point in its favor is that there is no way (currently) to make a prerequisite for illiteracy due to the fact that it is a) per-language, and b) a modifier on the language. Against, however, is that you can easily be illiterate in one language while being literate in another.

daliin:
Good points. Another possible argument in favor of the change is that it makes it easier for first time players who are a larger portion of the DFRPG crowd. If the book has a disadvantage, I would expect to see it in the library. (It does appear as an option on the Barbarian template.) 

But you know the full scope of GCS far better than I do (obviously!). Maybe this would muddy the waters in terms of libraries being fully compatible.

richardwilkes (approved):

PR #61, opened 2021-04-07
Changes preparing for Quintessence

Legendsmith:
This has the relevant Pyramid library changes for when Quintessence is integrated into the main branch of GCS.


Legendsmith:
Not sure why there's conflicts but this is otherwise ready to go. The old ``GURPS Magic Spells (Based on 10).spl`` needs to be deleted because it was made when there was that bug with difficulty so everything that/s VH is actually only H.

richardwilkes:
Hmm... this probably needs to be rebased on top of the current master.

Unrelated to that, however, is that this calls for the same type of support hit locations now has: a combo with pre-configured sets that you can load, and that are discovered via looking in well-known spots in the various Libraries you have defined (for Hit Locations, that's literally the directory "Hit Locations" in each of your libraries).

Legendsmith:
Well that's not how one rebases. Gonna redo this.

PR #63, opened 2021-04-13
Fix LT Clothing Costs & Default settings for Basic Set Languages

Legendsmith:
It's still doing the modifier ID thing, even when I edit in place.

### DIFF
--- a/Library/Basic Set/Basic Set Advantages.adq
+++ b/Library/Basic Set/Basic Set Advantages.adq
@@ -15491,7 +15491,7 @@
 				{
 					"type": "modifier",
 					"version": 1,
-					"id": "5b41b3bd-bf9d-4aa0-82c4-2f515b86ccb5",
+					"id": "ff972d9a-b925-4895-a489-f3d14999074d",
 					"disabled": true,
 					"name": "Native",
 					"reference": "B23",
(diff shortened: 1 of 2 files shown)


richardwilkes:
I see that... I forgot to look at that issue.

Why the change to language? To be correct, one selection for both spoken and written components should always be selected. GCS has no setting to enforce that, but the point of the current state was to make a valid entry without having to do anything further, as well as to encourage players to consider whether they needed to change something. Without that, too many people end up adding a 0-pt language (i.e. with no modifiers) to their sheet thinking they have full access to it.

Legendsmith:
Ah, well I noticed that change to language having accented/accented being enabled buy default had been done a while back. I found it a little annoying to have to disable those every time I wanted something different, but not a big deal. However recently one of my players remarked that "taking a native language gives you points back"; he thought that taking a native language was just a matter of ticking native, not realizing that the default was set to accented thus making it worth *negative* points in conjunction with the -6 'native' modifier. So I thought "this must be a mistake, time to fix it."

richardwilkes:
(sigh) Yeah, that's also a potential problem. For this particular case, it would probably be better for someone to take the time to just generate the different permutations, rather than using modifiers, since exactly one each for spoken & written should always be selected, no more, no less.

richardwilkes:
Anyway, happy to take the price changes... but I'd rather not take the language change, unless you want to take the time to generate the permutations and just get rid of the versions with modifiers.

Legendsmith:
Undid the language changes. Perhaps when custom stats or w/e are added we can have language & culture be their own box.

richardwilkes (approved):

PR #66, opened 2021-05-01
Fix broken one college prerequisites and rename abbreviated colleges to make it easier

fishface60:
A few spells had broken one college magery prerequisites which would permit with levels of one college magery in the wrong college because the clause that they required magery without a college limit was omitted.

This also led to discovering that one college prerequisites sometimes used abbreviated names for the colleges, which you wouldn't know if you'd picked the full name when adding the one college limit, and the college name is the only place the name of the college is advertised to the user.

In the process this also fixes a few bugs where the college names were inconsistent.

As an aside, if the next time there's a non-backwards-compatible change all the libraries could be re-saved to normalise to the canonical lower-case value names for spell difficulty, that would make putting together clean commits nicer.

### DIFF
--- a/Library/Magic/Magic Spells.spl
+++ b/Library/Magic/Magic Spells.spl
@@ -4747,7 +4747,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -4784,7 +4784,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -6619,7 +6619,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -6696,7 +6696,7 @@
 								},
 								"notes": {
 									"compare": "contains",
-									"qualifier": "one college (communication)"
+									"qualifier": "one college (communication & empathy)"
 								}
 							},
 							{
@@ -6733,7 +6733,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -7710,7 +7710,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -11454,7 +11454,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -12774,7 +12774,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -12825,7 +12825,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -12876,7 +12876,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -15144,7 +15144,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -18354,7 +18354,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -18421,7 +18421,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -20224,7 +20224,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -20361,7 +20361,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -22066,7 +22066,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -23208,7 +23208,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -23348,7 +23348,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -25106,7 +25106,7 @@
 				]
 			},
 			"categories": [
-				"Communication",
+				"Communication & Empathy",
 				"Technological"
 			]
 		},
@@ -26417,7 +26417,7 @@
 				]
 			},
 			"categories": [
-				"Communication",
+				"Communication & Empathy",
 				"Sound"
 			]
 		},
@@ -26545,7 +26545,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -26583,7 +26583,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -26620,7 +26620,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -28961,7 +28961,7 @@
 								},
 								"notes": {
 									"compare": "contains",
-									"qualifier": "one college (communication)"
+									"qualifier": "one college (communication & empathy)"
 								}
 							},
 							{
@@ -28998,7 +28998,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -29111,7 +29111,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
@@ -30373,7 +30373,7 @@
 								},
 								"notes": {
 									"compare": "contains",
-									"qualifier": "one college (communication)"
+									"qualifier": "one college (communication & empathy)"
 								}
 							},
 							{
@@ -30429,7 +30429,7 @@
 				]
 			},
 			"categories": [
-				"Communication"
+				"Communication & Empathy"
 			]
 		},
 		{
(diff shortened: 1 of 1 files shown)


Library/Magic/Magic Spells.spl
```diff
@@ -7900,7 +7904,7 @@
 			"name": "Coolness",
 			"reference": "M187",
 			"difficulty": "IQ/H",
-			"college": "Protection or Water",
+			"college": "Protection & Warning or Water",
```
richardwilkes, line 7907: This PR seems to be removing the cases where the author put more than one college in the college field... yet this seems to be an exception here. Why?

Library/Magic/Magic Spells.spl
```diff
@@ -10543,7 +10555,7 @@
 			"name": "Deflect Missile",
 			"reference": "M143",
 			"difficulty": "IQ/H",
-			"college": "Movement or Protection",
+			"college": "Movement or Protection & Warning",
```
richardwilkes, line 10558: Another case of retaining an "or"

Library/Magic/Magic Spells.spl
```diff
@@ -26100,7 +26136,7 @@
 			"name": "Metal Vision",
 			"reference": "M183",
 			"difficulty": "IQ/H",
-			"college": "Knowledge or Technological or Metal",
+			"college": "Knowledge or Technological",
```
richardwilkes, line 26139: another

Library/Magic/Magic Spells.spl
```diff
@@ -29828,7 +29868,7 @@
 			"name": "Plastic Vision",
 			"reference": "M183",
 			"difficulty": "IQ/H",
-			"college": "Knowledge or Technological or Plastic",
+			"college": "Knowledge or Technological",
```
richardwilkes, line 29871: another

Library/Magic/Magic Spells.spl
```diff
@@ -31866,7 +31906,7 @@
 			"reference": "M177",
 			"difficulty": "IQ/VH",
 			"tech_level": "",
-			"college": "Making & Breaking or Technological or Machine",
+			"college": "Making & Breaking or Technological",
```
richardwilkes, line 31909: again

Library/Magic/Magic Spells.spl
```diff
@@ -33741,7 +33781,7 @@
 			"name": "Resist Acid",
 			"reference": "M190",
 			"difficulty": "IQ/H",
-			"college": "Protection or Water",
+			"college": "Protection & Warning or Water",
```
richardwilkes, line 33784: again

Library/Magic/Magic Spells.spl
```diff
@@ -33816,7 +33856,7 @@
 			"name": "Resist Disease",
 			"reference": "M90",
 			"difficulty": "IQ/H",
-			"college": "Healing or Protection",
+			"college": "Healing or Protection & Warning",
```
richardwilkes, line 33859: more

Library/Magic/Magic Spells.spl
```diff
@@ -33941,7 +33981,7 @@
 			"name": "Resist Lightning",
 			"reference": "M196",
 			"difficulty": "IQ/H",
-			"college": "Air or Protection or Weather",
+			"college": "Air or Protection & Warning or Weather",
```
richardwilkes, line 33984: another

Library/Magic/Magic Spells.spl
```diff
@@ -34055,7 +34095,7 @@
 			"name": "Resist Poison",
 			"reference": "M91",
 			"difficulty": "IQ/H",
-			"college": "Healing or Protection",
+			"college": "Healing or Protection & Warning",
```
richardwilkes, line 34098: another

Library/Magic/Magic Spells.spl
```diff
@@ -34130,7 +34170,7 @@
 			"name": "Resist Radiation",
 			"reference": "M182",
 			"difficulty": "IQ/H",
-			"college": "Protection or Technological or Radiation",
+			"college": "Protection & Warning or Technological",
```
richardwilkes, line 34173: Another

Library/Magic/Magic Spells.spl
```diff
@@ -34169,7 +34209,7 @@
 			"name": "Resist Sound",
 			"reference": "M173",
 			"difficulty": "IQ/H",
-			"college": "Protection or Sound",
+			"college": "Protection & Warning or Sound",
```
richardwilkes, line 34212: Another

Library/Magic/Magic Spells.spl
```diff
@@ -34207,7 +34247,7 @@
 			"name": "Resist Water",
 			"reference": "M186",
 			"difficulty": "IQ/H",
-			"college": "Protection or Water",
+			"college": "Protection & Warning or Water",
```
richardwilkes, line 34250: Another

Library/Magic/Magic Spells.spl
```diff
@@ -35680,7 +35720,7 @@
 			"reference": "M177",
 			"difficulty": "IQ/VH",
 			"tech_level": "",
-			"college": "Knowledge or Technological or Machine",
+			"college": "Knowledge or Technological",
```
richardwilkes, line 35723: Another

Library/Magic/Magic Spells.spl
```diff
@@ -44483,7 +44531,7 @@
 			"name": "Teleport Shield",
 			"reference": "M170",
 			"difficulty": "IQ/H",
-			"college": "Gate or Protection",
+			"college": "Gate or Protection & Warning",
```
richardwilkes, line 44534: Another

Library/Magic/Magic Spells.spl
```diff
@@ -46038,7 +46086,7 @@
 			"name": "Umbrella",
 			"reference": "M185",
 			"difficulty": "IQ/H",
-			"college": "Protection or Water",
+			"college": "Protection & Warning or Water",
```
richardwilkes, line 46089: Another

Library/Magic/Magic Spells.spl
```diff
@@ -47092,7 +47144,7 @@
 			"name": "Warmth",
 			"reference": "M74",
 			"difficulty": "IQ/H",
-			"college": "Fire or Protection",
+			"college": "Fire or Protection & Warning",
```
richardwilkes, line 47147: Another

richardwilkes (commented):
Should have started the "review" rather than the individual comments earlier -- that's what I get for being half-asleep. Anyway, as mentioned earlier, there seems to be a mix of removing OR conditions in the college (which is likely for the best, since the field wasn't intended to be used that way originally) and leaving them in or even adding them in one or two cases. What was the intent here?

richardwilkes:
> As an aside, if the next time there's a non-backwards-compatible change all the libraries could be re-saved to normalise to the canonical lower-case value names for spell difficulty, that would make putting together clean commits nicer.

I've done that a few times. However, for spell difficulty, if you are using the current release, they are still in all uppercase. That has only been changed in the unreleased code, which should not be used to modify library files yet. All cases of lowercasing of that field in this PR should be reversed to be uppercase as is currently used by GCS to be safe.

richardwilkes:
Hmm... thinking some more about this, I think adding support for the college field to be an array of values might be the best approach. Would allow a single spell to belong to multiple colleges, which is what I believe is the case with the official rules for a subset of the current spells while allowing the code to apply the various checks to it correctly.

I'm in the middle of a sweeping set of changes right now, so not a bad time to do such a change. If you're feeling particularly energetic, you could split this into two parts: one for the basic fixes, minus removing / adding additional colleges that could be merged now, and one for the soon-to-be new code where multiple colleges would be supported. If that's not something you're up for, then I'd prefer to wait until the next release comes out and I've bumped the library version to 3.x.x along with the multiple college support.

fishface60:
> Should have started the "review" rather than the individual comments earlier -- that's what I get for being half-asleep. Anyway, as mentioned earlier, there seems to be a mix of removing OR conditions in the college (which is likely for the best, since the field wasn't intended to be used that way originally) and leaving them in or even adding them in one or two cases. What was the intent here?

Sorry, I should've been more clear in the commit messages.
The intent wasn't to change all of the spells that are part of multiple colleges to just one, that sort-of works as-is, since the feature in the Magery advantage for granting a bonus to colleges can filter by whether the college field contains the name of the college.
The intent was to remove the named colleges that seem to have been entered in error, mistaking subheadings in GURPS Magic for the aid of categorisation (e.g. Linking for Meta-Spells, Armor Enchantment, Radiation Technological spells) as colleges themselves.
The spell tables don't name them as colleges, and Thaumatology has suggestions for adding a Metal separately, which implies Magic isn't intending to define one.

> > As an aside, if the next time there's a non-backwards-compatible change all the libraries could be re-saved to normalise to the canonical lower-case value names for spell difficulty, that would make putting together clean commits nicer.
> 
> I've done that a few times. However, for spell difficulty, if you are using the current release, they are still in all uppercase. That has only been changed in the unreleased code, which should not be used to modify library files yet. All cases of lowercasing of that field in this PR should be reversed to be uppercase as is currently used by GCS to be safe.

Ah, that explains it. I thought I managed to catch all the diffs that changed the casing, thanks for pointing out that I missed some. I'll re-submit with that fixed.

> Hmm... thinking some more about this, I think adding support for the college field to be an array of values might be the best approach. Would allow a single spell to belong to multiple colleges, which is what I believe is the case with the official rules for a subset of the current spells while allowing the code to apply the various checks to it correctly.

The current approach sort-of works for applying spell bonuses and meeting prerequisites of having the right number of spells, since they filter by whether the prerequisite college is a substring of the spell's college.
With the removal of the Metal college (which would be given a bonus by one-college Magery for Meta-spells) there's no longer any ambiguity where a substring match causes problems, but Thaumatology has got Metal as a suggested variant college so it would be a useful change if a GM wanted to run a game with that.

> I'm in the middle of a sweeping set of changes right now, so not a bad time to do such a change. If you're feeling particularly energetic, you could split this into two parts: one for the basic fixes, minus removing / adding additional colleges that could be merged now, and one for the soon-to-be new code where multiple colleges would be supported. If that's not something you're up for, then I'd prefer to wait until the next release comes out and I've bumped the library version to 3.x.x along with the multiple college support.

I'll turn this PR into just the fixes and move the other changes to a new one.

fishface60:
Thank you for your patience. After writing a script to check my changes I discovered some more spells whose magery prerequisites omitted the note for not containing one college, others whose prerequisites didn't list every college the spell could be part of, and that I'd missed renaming the Light college entries where they required a number of spells in the college.

richardwilkes (approved):

PR #68, opened 2021-05-17
Quintessence Libraries from Pyramid 3/120 The Fifth Attribute for GCS 4.30

Legendsmith:


richardwilkes (changes requested):
Generally looks good (although I only skimmed the revised magic spell list as it is so long).

I'd prefer if you move the .gas file into a top-level "Attributes" directory in preparation for a future version of GCS that looks in that directory for pre-defined attribute sets to allow the user to select from.

Legendsmith:
Changes made as requested.

Legendsmith:
Made the changes. Just called the file Quintessence.gas, but was there some kind of naming scheme you had in mind?

richardwilkes (approved):

PR #69, opened 2021-05-18
More Attribute Presets

Legendsmith:
In preparation, I'll keep this here just so I don't spam PRs if there's more to be added.
Currently I've added Default, to reset back other changes,, and the popular Separate Will and Per from IQ.

### DIFF
--- a/Library/Attributes/Default.gas
+++ b/Library/Attributes/Default.gas
@@ -1,0 +1,299 @@
+{
+	"type": "attribute_settings",
+	"version": 2,
+	"name": "Default",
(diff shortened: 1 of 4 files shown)


richardwilkes:
I’m not against having Default there, but you do get the same effect in the Preferences panel by just hitting reset — albeit that resets *everything* in that tab, not just the attributes, so probably worthwhile. Same reason I included a “Humanoid” hit location table in the library, even though that one is also hard-coded into GCS so that it can reset back to it when asked.

richardwilkes (approved):

PR #71, opened 2021-05-19
Fixing all "whose name is <blank>" and similar issues

danielrab:
there are many places where something specifies compare mode but provides no string. Some of those should actually have something, many more should just be "is anything". This commit should fix all such instances. <br>
see issue https://github.com/richardwilkes/gcs_master_library/issues/70 for more details


richardwilkes (approved):
Looks good. Thanks for taking the time to clean that up.

PR #72, opened 2021-05-23
Programmatically fix one-college prerequisites in spell libraries

fishface60:
This is the same fix-ups from https://github.com/richardwilkes/gcs_master_library/pull/66 but programmatically applied to other Magic spell libraries.
There's also a lot of noise in reordering of prerequisites as a result of normalising them to coming first and being in order of no college then colleges in alphabetical order.
The easy changes have been split out into separate commits, but the last one is a mix of noise, renames and college removals.

If it helps to review I can provide the script, or take the last commit back and try to split it up further.


richardwilkes (approved):
Spot-checked the results and it looked good. Thanks.

PR #89, opened 2021-08-28
correct page refs to ACT1 bang skills

ilikerobots:
I believe the Action 1 Heroes "bang" skills e.g. _Assassin!_ are incorrectly referencing 13 instead of 23 (per my copy).


richardwilkes (approved):
I don’t have this particular book, so will take your word for it. Thanks for the PR.

PR #91, opened 2021-09-07
Removed Magery's invalid feature (+1/lvl spell lvl)

Deliphin:
I thought this +1 per Magery level to your spell level was valid, until a friend told me it wasn't. I looked at the Magery advantage, and double checked through Magic and lightly through Thaumatology. I can't find any mention of this anywhere in the actual books.
So unless someone else knows where this comes from, it should be removed. If I'm right that this is invalid, I'm surprised one of the most commonly used advantages has a mistake like this.
If this actually is valid and comes from a book other than Basic Set, then the advantage should mention this in the page reference section.


richardwilkes:
I'm afraid your friend is incorrect. This feature is clearly stated on page 6 of the Magic book: "Add your Magery to IQ when you learn spells. For instance, if you have IQ 12 and Magery 3, you learn spells as if you had IQ 15."

Deliphin:
..hm, I think I understand the confusion here, I think you're right.

Me and my friend interpreted this as affecting how learning spells work in GURPS, not casting. We thought this somehow affected the learning time for mages who are researching, not actually casting. But looking closer at it, I think it says learn because it's trying to say "Add your Magery to IQ on any spells you learned", to differentiate from spells you don't actually know and are reading out of a book, or something. It's a weird wording. I'll close the issue since you seem to be right.

PR #92, opened 2021-09-08
Traveller Alien Races I-V & Aliens Templates

Tarrith:
Upload to the master Library request for alien racial templates


dripton:
I imported your files into my GCS User Library and nothing blew up, so while there maybe be quibbles about exact values here or there, I think they're worth taking.

For the Traveller aliens see http://mygurps.com/pmwiki.php?n=Main.UpdatedTravellerRacialTemplates .  The author used to be the assistant GURPS line editor, so his 4E conversions should be about perfect, even though that's his personal page not an official SJG source.  So I would use his, for the aliens that he did.

For example, his Vargr is 16 points while yours is -12, mostly because he used Enhanced Move 1 with a 5% limitation for the early fatigue, while you used something much cheaper.

Tarrith:
Ah nice, I'll have a look at that :)

richardwilkes (approved):
I only quickly glanced at these. If additional changes are needed, a follow-up PR can be submitted.

Tarrith:
Cool thanks.
Yeah I'm sure there will be follow ups when people look at it and say...
"Oh no you messed that up, do it this way instead!"
;)

PR #115, opened 2021-12-20
a suggestion to standardize on the canonical damage abbreviation type…

Human:
…s for split DR

FWIW, I think it'd be easier on downstream consumers of GCS characters if the new split DR feature (thanks again for that!) used the same standard abbreviated damage types as weapons do.

Since some armor types have "DR vs piercing" which means `pi-`, `pi`, `pi+`, _and_ `pi++`, I was thinking that `pi*` might make sense to use in those cases. Thoughts? If you agree, I can follow up with some more equipment file edits to add this kind of split DR.

### DIFF
--- a/Library/Basic Set/Basic Set Equipment.eqp
+++ b/Library/Basic Set/Basic Set Equipment.eqp
@@ -7365,19 +7365,19 @@
 					"type": "dr_bonus",
 					"amount": -2,
 					"location": "torso",
-					"specialization": "crushing"
+					"specialization": "cr"
 				},
 				{
 					"type": "dr_bonus",
 					"amount": -2,
 					"location": "vitals",
-					"specialization": "crushing"
+					"specialization": "cr"
 				},
 				{
 					"type": "dr_bonus",
 					"amount": -2,
 					"location": "groin",
-					"specialization": "crushing"
+					"specialization": "cr"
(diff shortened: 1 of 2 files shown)


richardwilkes:
I'm torn on this. The abbreviation "cr", while standard for GURPS, isn't necessarily as readable as "crushing". Since this only shows up in the tooltip and the editor, space isn't really an issue.

I'd like to hear some opinions from others before I switch away from the long form for these fields. Want to start a discussion on the Discord and see what others think?

Human:
@richardwilkes I'd be happy to! I don't know if you have much interaction with the Foundry VTT GURPS Game Aid folks, but they would be consumers of this new split DR feature as well. Any way to loop them in, too?

richardwilkes:
I typically read the GURPS channel on the Foundry Discord, my own Discord (obviously!), and the GCS & Foundry-related channels on Mook's Discord. Any or all of them are fine with me.

richardwilkes:
IIRC, the discussions on Discord were fairly limited, but mostly biased in favor of having the longer, more readable names since these only show up in tooltips. Given that, I'm closing this out. If I've mis-remembered or there are further arguments to be made in favor of making the change to the abbreviated variants, please re-open.

PR #116, opened 2021-12-20
give Dungeon Fantasy 1 Equipment.eqp's Ironskin Amulet the same stats…

Human:
… as RPG/Dungeon Fantasy RPG Equipment.eqp's entry (which seems to be the correct interpretation)


richardwilkes (approved):
LGTM

PR #117, opened 2021-12-25
Added Conditional Modifiers loop

crnormand:
Although this will hopefully soon be obsolete... I promised to update the export function to include the Conditional Modifiers.


richardwilkes:
Looks like there are a number of unnecessary whitespace changes, but if you're OK with that, I'll merge it.

crnormand:
Yeah.  Not certain where those came from, but yes, they won't affect our import.  And thank you.

PR #129, opened 2022-02-17
removed duplicate entries

angelforest:


Library/Power Ups/Power Ups Advantages.adq
```diff
@@ -9712,234 +9712,6 @@
 				"Talent"
 			]
 		},
-		{
-			"type": "advantage",
-			"id": "916c9bae-7bad-4225-8f7a-046665599ed4",
-			"name": "Talent (Chi Talent)",
```
richardwilkes, line 9718: I noticed that this version didn't have the superfluous "Reaction Bonus: None" note... which made me then realize that all the other Talents that actually *do* have a reaction bonus were either done before you could actually specify reaction bonuses in GCS, or were just done wrong. Any chance you'd be willing to go through and adjust them as part of this PR?

Library/Tactical Shooting/Tactical Shooting Skills.skl
```diff
@@ -720,26 +700,6 @@
...
-			"name": "Targeted Attack (Chiks in torso armor)",
-			"reference": "TS45",
-			"difficulty": "h",
-			"points": 2,
-			"limit": -4,
-			"default": {
-				"type": "skill",
-				"name": "Guns",
-				"specialization": "@Guns@",
-				"modifier": -8
```
richardwilkes, line 736: Again, I don't have the relevant books, so can't say for certain, but this appears to be a special sub-case (torso armor vs other armor) with a different modifier. Is that not the case?

angelforest: Yes, but both of them were entered twice. With this change only one copy of each technique is present.

Library/Power Ups/Power Ups Advantages.adq
angelforest: Sure, I'll take a look.

angelforest: I changed reaction text to make it clear to whom it applies.
I'm not sure if there is an easy way to fix conditional bonuses without manually rewording every single one of them, and in my opinion it's too much effort when "+1: Bonus to <thing>" is reasonably clear even if a bit unnatural. Btw, mangled Survivor text was probably someone's attempt to shorthand the full description.

angelforest:
I apologize for the mess in notifications, but it should be fine now.

richardwilkes (approved):

PR #133, opened 2022-02-23
Added extra references to traits in Basic Set and Powers

angelforest:
This should make it easier to find relevant information when working with multiple books.

I also added modifiers and a variant of Modular Abilities that I noticed were missing.


Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -1564,7 +1564,7 @@
 					"notes": "@One species: House Cats, Macaws, etc.@"
 				}
 			],
-			"reference": "B40",
+			"reference": "B40,48",
```
richardwilkes, line 1567: I don't believe this notation is supported. You can use "B40,B48", but even then, the result will be less than useful, since the current options are to open the first one or to open them all -- but opening them all will result in the last one only in this case, since they refer to the same book.

Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -7280,7 +7280,7 @@
 					"affects": "total"
 				}
 			],
-			"reference": "B51",
+			"reference": "B51,48",
```
richardwilkes, line 7283: Same comment as above... not a valid format

Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -21558,7 +21919,7 @@
 					"notes": "@Small Subset: Evergreens, Medicinals@"
 				}
 			],
-			"reference": "B75",
+			"reference": "B75,48",
```
richardwilkes, line 21922: Another case

Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -26389,7 +26750,7 @@
 					"affects": "total"
 				}
 			],
-			"reference": "B88",
+			"reference": "B88,48",
```
richardwilkes, line 26753: Another

richardwilkes (approved):

PR #135, opened 2022-03-02
Added Gun Fu Reference Data

Calmquist:
I added Advantages, Skills, Techniques and Equipment from Gun Fu. I also added Limitations to the Gunslinger from the Basic Set listed in Gun Fu.


Library/Basic Set/Basic Set Advantages.adq
```diff
@@ -9504,6 +9504,49 @@
 			"name": "Gunslinger",
 			"mental": true,
 			"base_points": 25,
+<<<<<<< HEAD
```
richardwilkes, line 9507: This isn't valid... part of the merge conflict text

Calmquist: I must have missed that part. It should be fixed now.

richardwilkes (approved):

Issue #109, opened 2021-12-10
TL6 Geiger Counter is missing from the High Tech Equipment list


tjed:
TL6 Geiger Counter is missing from the High Tech equipment list (the TL8 is listed, though) 

This was listed on page 49 of the GURPS High Tech book

Bizarrely, the After the End equipment list does have the TL6 Geiger Counter (with a cost modifier)

richardwilkes:
I've transferred this issue and previous ones to the gcs_master_library repo. Please make sure you post into the correct repo in the future.

Calmquist:
I fixed this in #136.

PR #136, opened 2022-03-03
Added TL6 Geiger Counter.

Calmquist:
I added the TL 6 Geiger Counter. 

I am not sure how the "TsNIITochMash VSS Vintorez, 9x39mm" got changed since I didn't remove the armor divisor, but I verified that the change is correct. The armor divisor is for the "TsNIITochMash AS Val, 9x39mm" which is pi-.


Calmquist:
Resolves #109

richardwilkes (approved):

PR #137, opened 2022-03-05
Added Dwarven and Dverger ancestries.

Calmquist:
I added Dwarf and Dvergr ancestries. I adjusted the height, weight and age based on GURPS Fantasy and Dungeon Fantasy RPG descriptions. For the Dwarf, I based the skin and hair on Human.

For Dvergr, I based the descriptions on Norðlonder Folk using the Dwarf ancestry as a template to fill any gaps.

I created initial lists of names that "sound dwarf," but they are ultimately fairly arbitrary.


richardwilkes:
I don't think the names list have enough options to not have a lot of repetition. Lacking something better, though, we can live with that. I've not yet had a spare moment to try this out. Once I do, assuming no glaring issues, I'll merge it.

Calmquist:
The names were definitely an initial draft. They can certainly be improved.

PR #138, opened 2022-03-13
feat: adds the 'Social Engineering' library

Coppini:
Adds `Skills`, `Advantages` and `Modifiers` from [Social Engineering](http://www.sjgames.com/gurps/books/socialengineering/).

Resolves https://github.com/richardwilkes/gcs_master_library/issues/110

### DIFF
--- a/Library/Social Engineering/Social Engineering Modifiers.adm
+++ b/Library/Social Engineering/Social Engineering Modifiers.adm
@@ -11,7 +11,7 @@
 			"cost_type": "percentage",
 			"cost": -20,
 			"affects": "total",
-			"notes": "Choose aspect"
+			"notes": "@Aspect@"
 		}
 	]
 }
--- a/Library/Social Engineering/Social Engineering Modifiers.adm
+++ b/Library/Social Engineering/Social Engineering Modifiers.adm
@@ -1,0 +1,17 @@
+{
+	"type": "modifier_list",
+	"version": 2,
+	"id": "981785b6-2410-47bb-ad46-932ac21e6a50",
+	"rows": [
+		{
+			"type": "modifier",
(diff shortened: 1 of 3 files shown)


Library/Social Engineering/Social Engineering Modifiers.adm
```diff
@@ -0,0 +1,17 @@
...
+	"rows": [
+		{
+			"type": "modifier",
+			"id": "39ce3558-7397-40b5-8cf8-1d51b7efb9da",
+			"name": "Aspected",
+			"reference": "SE77",
+			"cost_type": "percentage",
+			"cost": -20,
+			"affects": "total",
+			"notes": "Choose aspect"
```
richardwilkes, line 14: Typically this is done using replaceable text, i.e. something like "@Aspect@", which will then prompt the user to enter in a value when they try to apply it to a character sheet. You can put these in the name or notes field.

richardwilkes (commented):
Generally looks good. Had one comment about the possible use of replaceable text, which may apply elsewhere as well -- I didn't look closely enough to identify other possible spots, though.

Library/Social Engineering/Social Engineering Modifiers.adm
richardwilkes: @Coppini were you planning to update this?

Coppini:
```suggestion
			"notes": "@Aspect@"
```
Sorry, hadn't noticed your comment before.

Coppini: Updated 👍🏻

richardwilkes (approved):

PR #150, opened 2022-06-29
StarWars v1

Nikofrom94:
Hi Richard,
Here are some library files for Starwars based on Darklord Azagthot's work (of the 16/3/2016 edition).
- Advantages : quite complete
- Force Habilities : complete
- Skills : only Force related skills
- Equipment : very very limited items
- Templates - Lightsaber forms : complete
- Templates - Equipments : tentative for two armors
- Templates - Species : Abyssin to Chevin (15% ?)

Will push updates from time to time.


richardwilkes:
Looks reasonable, if incomplete, to me. Seems worth adding to the library. Thanks.

It would be better if there were page references to wherever this data was being sourced from, but will merge as-is.

PR #162, opened 2022-08-02
Add Pyramid 3/99

Calmquist:
I added Necrokinesis and Soul Reapers from Pyramid 3/99. I also added Expert Skill (Hopology) to Martial Arts since it was missing.


Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -1049,6 +1049,18 @@
 				"Technique"
 			]
 		},
+		{
+			"type": "skill",
+			"id": "f5c71028-8918-4424-bd42-f1eb53a16604",
+			"name": "Expert Skill (Hopology)",
+			"reference": "B193,MA:56",
```
richardwilkes, line 1056: The page ref of MA:56 doesn't look right. Should this be MA56 instead?

Calmquist: I fixed that.

richardwilkes (approved):

PR #164, opened 2022-08-12
Fixed some adavantages and added Additional Powers.

Calmquist:
Fixed "weapons" for TK Bullet and Lightning to use the psionic skill rather than Innate Attack.
Fixed Mental Stab skill to require Mental Stab advantage rather than Mental Blow.
Added Additional Powers from Psionic Powers pp. 71-76.

### DIFF
--- a/Library/Psionics/Psionics Advantages.adq
+++ b/Library/Psionics/Psionics Advantages.adq
@@ -5,6 +5,69 @@
 	"rows": [
 		{
 			"type": "advantage",
+			"id": "84aafd42-e3cc-4a2d-a717-7cf189907afc",
+			"name": "Enhanced Power Defence",
+			"mental": true,
+			"exotic": true,
+			"modifiers": [
+				{
+					"type": "modifier",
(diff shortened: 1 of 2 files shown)


Library/Psionics/Psionics Advantages.adq
```diff
@@ -3,6 +3,69 @@
 	"version": 2,
 	"id": "4a1c9802-f8dc-43bc-a1df-3750b319427b",
 	"rows": [
+		{
+			"type": "advantage",
+			"id": "84aafd42-e3cc-4a2d-a717-7cf189907afc",
+			"name": "Enhanced Power Defence",
```
richardwilkes, line 9: Does the book actually spell this "Defence" and not "Defense"?

Calmquist: I didn't actually add this. I merely alphabetized the list. I checked the book, and it is indeed simply a typo, so I fixed it.

richardwilkes: Thanks

richardwilkes (approved):

PR #166, opened 2022-08-17
Add licenses and certificates from GURPS Traveller Far Trader.

dripton:
These should be identical to the ones on GURPS Traveller Far Trader
pages 83-85, with some small changes to remove references to skills in
GURPS 3E that are no longer found in 4E.  Shipbuilding has been merged
into Engineer (Starships).  Shipmaster is now Shiphandling (Starship).
This removes the distinction between Limited and Unlimited Master's and
Mate's Licenses, so now there is just one type of Master's License and
one type of Mate's License.

Note that a few of the licenses require a certain number of points
spread across multiple skills, or multiple specializations of a skill,
of skills, which are not supported by the current GCS prerequisite
system, so for those licenses that part of the requirement is just
written in the notes rather than actually enforced.  Players and GMs
should manually check that characters actually fulfill any requirements
listed in the notes.

### DIFF
--- a/Library/Traveller/Traveller Standard Licenses and Certificates.adq
+++ b/Library/Traveller/Traveller Standard Licenses and Certificates.adq
@@ -1,0 +1,690 @@
+{
+	"type": "advantage_list",
+	"version": 2,
+	"id": "5c442304-8fb6-498f-b96c-7b0bb989048c",
+	"rows": [
+		{
+			"type": "advantage",
+			"id": "f218d1ef-62ba-4e37-8744-89e7a4908bac",
+			"name": "Certified Medical Technician",
+			"mental": true,
+			"reference": "GTFT84",
+			"calc": {
+				"points": 0
+			},
+			"prereqs": {
+				"type": "prereq_list",
+				"all": true,
+				"prereqs": [
+					{
+						"type": "skill_prereq",
+						"has": true,
+						"name": {
+							"compare": "is",
+							"qualifier": "Electronics Operation"
+						},
+						"level": {
+							"compare": "at_least",
+							"qualifier": 12
+						},
+						"specialization": {
+							"compare": "is",
+							"qualifier": "Medical"
+						}
+					},
+					{
+						"type": "skill_prereq",
+						"has": true,
+						"name": {
+							"compare": "is",
+							"qualifier": "First Aid"
+						},
+						"level": {
+							"compare": "at_least",
+							"qualifier": 12
+						}
+					},
+					{
+						"type": "skill_prereq",
+						"has": true,
+						"name": {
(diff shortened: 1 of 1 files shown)


richardwilkes:
Shouldn't these all be equipment and not advantages?

Unrelated to that, the file name should be changed to have spaces at appropriate points. There is no need to avoid such things.

dripton:
I think they are advantages.  The license or certificate indicates that you have certain skills; it's not a magic item that gives you certain skills.

I will modify the commit to put spaces in the filename.

richardwilkes:
In real life, wouldn't all of these be actual documents, though? If the concern is that you'd have to enter them in again, I think it should be straight-forward for me to convert these directly to equipment.

dripton:
I think "Medical Doctor" is an advantage that you get for free if you have enough points in the right skills, went to the right school, passed the right government tests, paid the right fees, etc.  Whether you have a piece of paper that says you are one is kind of secondary, because if the paper is destroyed, it can be replaced.  The source of truth is a series of records in a series of databases, not the diploma or certificate or clothing that you may have been given as a symbol of the achievement.

Similar advantages would be Legal Enforcement Powers or Clerical Investment.  There may be physical symbols that go with them, but you are a cop or a priest even if you lose your badge or vestments.

But the place where I agree with you is if it's forged.  If all you have is a fake cert that you carry around, not backed up by real database entries somewhere, then that's equipment rather than an advantage.  Then again, a Gizmo is equipment that's tracked like an advantage, so there's precedent.

richardwilkes:
While I agree that the piece of paper a certificate is printed on isn't terribly important, that's still the way it exists in real life. You don't just say you're a doctor, you have a physical certificate as well. Yes, there is plenty of electronic records that go along with that certification as well... but the primary thing *you* have is the document stating you've achieved this milestone.

Anyway, if you're dead-set against the change to equipment, I'll let it go, since I don't use this myself anyway.

re: Gizmo... yeah, that's a very special case, and (from my point of view) is more like a magic ability than anything else. It's another thing I don't use in my games, since I prefer realism over hand-waving such details. Signature Gear is another one, although in that case, I'd have two entries: one for the advantage that specifies what the signature gear is, and one for the actual piece of equipment.

dripton:
Just to muddy the waters, I checked GT: Far Trader and some Journal of the Travellers' Aid Society posts, and the characters that have certs or licenses have them in a separate category, neither equipment nor advantages.  But clearly it's not worth making a whole new category in GCS for this tiny thing, so I think advantages are the best fit, because these are mostly immutable parts of a character rather than something you buy and sell and find and lose.

richardwilkes:
OK, fair enough

PR #167, opened 2022-08-20
Added Shield-Bearer - PY118, Mystic Swordman - PY36, Demonologist - DF9

jcklemme:


richardwilkes:
Let me know when you have the final version of these changes in place, so I can review it all at once.

jcklemme:
> 

This is all I'm planning on working on for the week.

richardwilkes:
The college for the demonologist spells doesn't seem right. Shouldn't it be something like "Demonology" instead?

jcklemme:
> The college for the demonologist spells doesn't seem right. Shouldn't it be something like "Demonology" instead?

It doesn't actually flat out say what the college name is in DF9, and I scratched my head about it for a while and decided to go with "Demonological" since the abilities are classified as Demonological.  I was kind of "computer braining" it and wanted to match as closely to the nearest thing.  If that's not right or I missed something you saw I'd believe you lol

PR #180, opened 2022-10-18
Update HTML - PC

CorwinBurton:
Fixes the css grid "traits" box - new exporter swapped names from advantages Adds @ values to Skills Loop


richardwilkes (approved):
Looks good. Thanks.

PR #183, opened 2022-10-27
Add new gcs 5.3 templates to the Traveller career templates

dripton:
There are some edge cases that don't work perfectly, like advantages and disadvantages with variable point values.  But it's a good start.

### DIFF
--- a/Library/Traveller/Templates/Academician.gct
+++ b/Library/Traveller/Templates/Academician.gct
@@ -3730,181 +3730,6 @@
 									"modifier": -5
 								}
 							]
-						},
-						{
-							"id": "8e681f08-b2d1-4b18-9249-67b6ee638aea",
-							"type": "skill_container",
-							"open": true,
-							"children": [
-								{
-									"id": "13262d1a-7e56-4f4d-b0d0-7cf53b8e3cf8",
-									"type": "skill",
-									"name": "Computer Operation",
-									"reference": "B184",
-									"tags": [
-										"Everyman",
-										"Scholarly",
-										"Technical"
-									],
-									"tech_level": "",
-									"difficulty": "iq/e",
-									"points": 2,
-									"defaults": [
-										{
-											"type": "iq",
-											"modifier": -4
-										}
-									]
-								},
-								{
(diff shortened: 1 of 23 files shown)


richardwilkes:
I've only looked at the first one (Academician) so far and noticed that you have a "Pick 2" for Primary Skills where there are a bunch of 4 point skills... but also a container that has 8 points worth inside it. Thanks to the way things work, that container is treated as a single entry, so if you apply this template and select the container as one of your choices, you would still need to select a 4 point skill to go along with it. Is that what you actually intended?

richardwilkes:
Note: I was thinking that maybe that "Secondary Skills" container was intended to be up a level and therefore not part of the "Pick 2" requirement.

dripton:
Bug, secondary skills is supposed to be outside primary skills.  I went through the other templates and didn't see the same bug anywhere else, but it's always possible.  Will submit a fix.

richardwilkes:
I also noticed in Bureaucrat that you have a "2 extra points in Primary or Secondary skills" and it appears that the intent is that they get added to existing skills. That, unfortunately, won't actually do what you're intending. You'll end up with the skill repeated. For example, if I chose Administration as one of the extras, I'd have Administration (4 points) and Administration (1 point) in my list when done.

dripton:
Yeah, I don't see a better way to handle that within the current code.  If you pick different skills then it should work, but if you pick the same skill then some manual editing will be needed to clean things up later.  Do you see a way?

richardwilkes:
No, I don't. It is worth creating an issue for the GCS repo to have some way to support that sort of thing. I do already intend to allow some sort of mechanism where points can be adjusted in the picker (i.e. where you say "pick 8 points worth" and you can then adjust any of the selection's actual point values to meet that criteria rather than relying on them being pre-set to specific values... of course, I'll want that to be a specific option to allow/disallow, since it isn't always appropriate.

richardwilkes:
Anyway, I'm going to merge this. Thanks!

dripton:
I will make an issue for this in the gcs repo.

PR #184, opened 2022-10-29
Add template choices to DFRPG class templates

dripton:


Library/Dungeon Fantasy RPG/Classes/Barbarian.gct
```diff
@@ -2065,6 +2117,15 @@
...
 								},
+								{
+									"id": "15608554-8633-4b15-8e27-a821ccbd640b",
+									"type": "trait_container",
+									"open": true,
+									"name": "Trait Container",
+									"calc": {
+										"points": 0
+									}
+								},
```
richardwilkes, line 2128: What's the point of this empty trait container?

richardwilkes:
Since there is currently no way for users to select the appropriate modifiers for things like Appearance in the template picker, it would likely be better if you created separate instances of each applicable base state (in Barbarian case, it looks like you already narrowed the modifier choices down to two, so create two entries instead, each with just the one modifier enabled). This will cause the trait to be priced appropriately, allowing them to actually meet the criteria.

Library/Dungeon Fantasy RPG/Classes/Barbarian.gct
dripton: Typo, fixed

dripton:
I agree that would be better.  So far, I have just made minimal changes, finding places where there was a choice and adding the new choice templates.  I can take another pass and try to find places where a trait should be fragmented into multiple traits.

dripton:
I split 0-point Wealth, Appearance, and Languages into fixed point versions.  One issue is that Languages could be taken more than once; I don't see a good way to represent that in the current system.  (I could do Language 1, Language 2, etc.  But how many is a good limit?)

richardwilkes:
Sense of Duty looks like another one that should be expanded to the two possibilities listed (in the Barbarian template, anyway).

dripton:
I split up the Barbarian Sense of Duty.  Looks like other classes with two kinds of Sense of Duty already had them split.

richardwilkes:
I see a few other things that could be split out, but I think I'll just quit nit-picking at merge. This can always be followed up with refinements.

PR #193, opened 2022-11-25
Add missing Martial Arts techniques

dec-00:
- Added various missing techniques

- Added Targeted Attack (Weapon)
- Added grappling versions of Targeted Attacks, as they have different defaults and maximums
- Put Targeted Attacks into a container

- Removed redundant prerequisites from some techniques

- Fixed some tags

### DIFF
--- a/Library/Martial Arts/Martial Arts Skills.skl
+++ b/Library/Martial Arts/Martial Arts Skills.skl
@@ -17,21 +17,7 @@
 				"name": "Acrobatics",
 				"modifier": -6
 			},
-			"limit": 0,
-			"prereqs": {
-				"type": "prereq_list",
-				"all": true,
-				"prereqs": [
-					{
-						"type": "skill_prereq",
-						"has": true,
-						"name": {
-							"compare": "is",
-							"qualifier": "Acrobatics"
-						}
-					}
-				]
-			}
+			"limit": 0
 		},
 		{
 			"id": "5f7627e8-c497-4839-8d7d-776b088f6686",
@@ -110,21 +96,7 @@
 				"type": "skill",
 				"name": "@Grappling or Melee Weapon Skill@"
 			},
-			"limit": 4,
-			"prereqs": {
-				"type": "prereq_list",
-				"all": true,
-				"prereqs": [
-					{
-						"type": "skill_prereq",
-						"has": true,
-						"name": {
-							"compare": "is",
-							"qualifier": "@Grappling or Melee Weapon Skill@"
-						}
-					}
-				]
-			}
+			"limit": 4
 		},
 		{
 			"id": "d67ea22a-d9dd-49dc-8bb5-0a02fbcf3e97",
@@ -144,21 +116,7 @@
 				"name": "@Cloak or Melee Weapon Skill@",
 				"modifier": -2
 			},
-			"limit": 0,
-			"prereqs": {
-				"type": "prereq_list",
-				"all": true,
-				"prereqs": [
-					{
-						"type": "skill_prereq",
-						"has": true,
-						"name": {
-							"compare": "is",
-							"qualifier": "@Cloak or Melee Weapon Skill@"
-						}
-					}
-				]
-			}
+			"limit": 0
 		},
 		{
 			"id": "a5734eee-5946-4ef1-aedf-aa674eefc5b4",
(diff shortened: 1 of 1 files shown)


Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -2280,104 +2313,264 @@
...
+				"name": "Wrestling",
+				"modifier": -2
+			},
+			"limit": 4
+		},
+		{
+			"id": "12ba82ac-9567-49b4-9358-ad78fd632b02",
 			"type": "technique",
-			"name": "Low Fighting",
+			"name": "Mounted Shooting",
```
Calmquist, line 2458: Mounted Shooting does have Riding or a vehicle-operation skill as a prerequisite.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -1884,6 +1796,27 @@
 				"name": "@Melee Weapon Skill@",
 				"modifier": -5
 			},
+			"limit": 0
+		},
+		{
+			"id": "a2df4d38-06b2-42a9-b529-d80b969dd90e",
+			"type": "technique",
+			"name": "Horse Archery",
```
Calmquist, line 1804: Horse Archery does have a Riding prerequisite.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -3017,241 +3229,466 @@
...
+								"Weapon"
+							],
+							"difficulty": "h",
+							"points": 2,
+							"default": {
+								"type": "skill",
+								"name": "@Skill@",
+								"modifier": -3
+							},
+							"limit": 0
```
Calmquist, line 3336: Limit should be -1.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -3017,241 +3229,466 @@
...
+								"Weapon"
+							],
+							"difficulty": "h",
+							"points": 2,
+							"default": {
+								"type": "skill",
+								"name": "@Skill@",
+								"modifier": -2
+							},
+							"limit": 0
```
Calmquist, line 3356: Limit should be -1.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -3017,241 +3229,466 @@
...
```
Calmquist, line 3376: Limit should be -1.

Calmquist (changes requested):
A few techniques are missing prerequisites, but it is good overall. I see no problem with removing the redundant prerequisites.

Library/Martial Arts/Martial Arts Skills.skl
dec-00: I did include this.
Line 1830

Library/Martial Arts/Martial Arts Skills.skl
dec-00: I wasn't quite sure what to do in this case.
I assumed there to be way too many Vehicle skills to be practical to list them all in the prerequisites. For example, I assumed that each specialization of Driving would be counted as a separate skill in GCS (fortunately not the case).
I just checked out GURPS Skill Categories and there's only 14 actual Vehicle skills. A lot, but definitely doable. So I'll just get on that real quick.

dec-00: So it turns out even less of those were actually involved with operating a vehicle, so that's fortunate.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -1884,6 +1796,27 @@
...
+			"reference": "MA74",
+			"tags": [
+				"Animal",
+				"Combat",
+				"Ranged Combat",
+				"Technique",
+				"Weapon"
+			],
+			"difficulty": "h",
+			"points": 1,
```
Calmquist, line 1814: The first level of a hard technique should be 2 points.

Library/Martial Arts/Martial Arts Skills.skl
```diff
@@ -2280,104 +2313,264 @@
...
+				"Animal",
 				"Combat",
-				"Melee Combat",
+				"Ranged Combat",
 				"Technique",
 				"Weapon"
 			],
 			"difficulty": "h",
-			"points": 2,
+			"points": 1,
```
Calmquist, line 2468: The difficulty is hard, so the first level should be 2 points.

Calmquist (changes requested):
That looks good, but some hard techniques were switched to 1 point for some reason. The first level should be 2 points.

dec-00:
> That looks good, but some hard techniques were switched to 1 point for some reason.

I honestly have no idea why this happened. I did all the editing within GCS itself (with the exception of the commit I just made) so I couldn't see the point values.

With some testing, it seems that techniques default to 1 point even if you make them Hard, usually. But it's inconsistent. Many of the new techniques I added were correctly at 2 points. 

To save time when adding some techniques, I'd duplicate another technique and edit its values to match the new technique. Sometimes this would end up changing the point value to be wrong, like how you pointed out when changing the copy of Low-Fighting to Mounted Shooting; however, sometimes the opposite would happen, and the point value would update to 2 after changing the difficulty to Hard (Lower-Body Arm Lock to Noogie (Karate) as an example).

I assume this is a bug with GCS, but I wouldn't have a clue what exactly causes which behavior.

Calmquist (approved):

richardwilkes:
Please create an issue on the GCS issues list with steps you took that resulted in bad behavior, @dec-00

PR #199, opened 2023-01-02
Various amendments

ethorad:
Added intimidation bonuses for poor appearance (B202)

Removed No Fine Manipulators limitation from Extra Hit Points

Clarified the "Reputation" and "Status" traits to split good/bad

Clarified the reaction modified on excommunicated to only apply to people following your faith

### DIFF
--- a/Library/Basic Set/Basic Set Advantages.adq
+++ b/Library/Basic Set/Basic Set Advantages.adq
@@ -1735,11 +1735,20 @@
 					"id": "cfef51b4-8526-47be-bb28-49c4e7db46b5",
 					"type": "modifier",
 					"name": "Hideous",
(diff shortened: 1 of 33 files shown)


ethorad:
Racial templates from the 3e Fantasy Folk pdf, converted using the 3e to 4e update guidelines.  I used FF as the code to link to the pdf since that wasn't currently used and fits with the use of FFE for the 4e Fantasy Folk: Elves

richardwilkes:
I've not had the time to go over anything other than the Basic Set changes you've made, which look good to me. Since no one else in the community is stepping up to take a look, I'm going to go ahead and just approve them. They can always be adjusted later. Thanks for the contribution!

PR #203, opened 2023-01-18
Adding some advantage modifiers, correcting some racial templates

ethorad:
Advantages:
- good/bad reputation: reaction modifiers
- chameleon: extended enhancement has cost of +20%
- clueless: clarified condition for reaction penalty
- cold blooded: HT modifiers for temperature effects
- dependency / dependent correction I think
- discriminatory hearing / smell / taste: added conditional bonuses
- nictating membrane: HT modifier to eye damage
- vow: change tag to mental
- weirdness magnet: added reaction penalty

Race templates (from the Fantasy Folk unless noted):
- vampire (basic set): Added dominance and tweaked weakness for the right points
- ghoul: tweaked how the degenerate if living among orcs was dealt with
- onocentaur: moved its racial skill into the container
- giant, orc: changed container from alternative ability to group

### DIFF
--- a/Library/Basic Set/Basic Set Advantages.adq
+++ b/Library/Basic Set/Basic Set Advantages.adq
@@ -2239,6 +2239,101 @@
 			}
 		},
 		{
+			"id": "bb686b77-de17-499e-b91f-a13b63368211",
+			"type": "trait",
+			"name": "Bad Reputation",
+			"reference": "B26,MA54",
+			"tags": [
+				"Disadvantage",
+				"Social"
+			],
+			"modifiers": [
+				{
+					"id": "4ab6e069-8e13-4a45-82bd-e07fbebf17a7",
+					"type": "modifier",
+					"name": "People Affected",
+					"reference": "B27",
+					"notes": "Almost everyone",
+					"cost": 1,
+					"cost_type": "multiplier",
+					"disabled": true
+				},
+				{
+					"id": "cd068eef-48af-4246-b1f4-c634f7c24a85",
+					"type": "modifier",
+					"name": "People Affected",
+					"reference": "B27",
+					"notes": "Almost everyone except @large class of people@",
+					"cost": 0.67,
+					"cost_type": "multiplier",
+					"disabled": true
+				},
+				{
+					"id": "d65bc4bc-7b2c-448f-bac1-09f70431fb57",
+					"type": "modifier",
+					"name": "People Affected",
+					"reference": "B27",
+					"notes": "@Large class of people@",
+					"cost": 0.5,
+					"cost_type": "multiplier",
+					"disabled": true
+				},
+				{
+					"id": "8ecfba87-4088-4469-baa7-9c83d0be8f02",
+					"type": "modifier",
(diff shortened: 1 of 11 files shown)


smithkm:
You've changed the order of some of the traits while also changing their content which makes it hard to see exactly what you changed in each individual trait.  If the re-ordering was intentional it would be better off being a separate commit.

ethorad:
That's probably what I did with dependent / dependency by accident. In my notes there wasn't a change to those traits but the diff showed them as swapping. 

Will have a look when I get back home and see if they should be swapped back and can update the pull request. 

This is my first time with git so please let me know if there's something I should be doing differently. At the moment I'm conscious I'm putting things into one big request, because I'm forking the master and making changes but it will only let me do one fork. 

Thanks

smithkm:
The "fork" is your own personal copy of the git repository within the GitHub service.  You can have many "branches" within the "fork".  If you look at my PR (#206)  It's pulling from `smithkm:ma-disadvantages` (The `ma-disadvantages` branch within my fork) to `richardwilkes:master`.  I was planning to do a similar change to what you did to reputation as a different branch in my fork so I could have made them separate PRs then I saw yours was already doing that.

I'm not sure exactly which tools you are using locally.  I'm used to using the command line tools but I know a lot of people prefer graphical Git clients.  In your local repository you can create branches then "push" them to your fork in GitHub.

It's been quite a while since I've contributed anything to GCS myself  (not since before the switch from Java to Go) so maybe Richard doesn't care.  Feel free to disregard my suggestion but making the diff "nice" will hopefully make it easier for Richard which means your PR is more likely to be accepted more quickly, and it chews up less time he could be writing new features instead.

ethorad:
That does look cleaner. 

So if I fork the master once. Then for each related set of changes, create a branch of my fork. Would i then push each branch to the master, or push the branches to my fork and then the fork to the master?

I'm using the graphical windows interface as that looked more user friendly than command prompt. 

I might remove this push when I get home and try to do it as something easier to read. And see what's going on with the dependency / dependent traits

smithkm:
A typical git workflow with GitHub would be

* Switch to your local `master` branch
* Pull from the `master` branch in the "upstream" repo (Richard's in this care) to make sure it's up to date
* Create a new branch 
* Make a change meaningful, stand alone change
* Commit the change with a descriptive message
* Repeat until all the changes needed for what you are doing are made
* "push" the branch to your fork
* Make a pull request from your branch to the upstream `master`.

Then start over again from `master`, but be aware that if you try to make parallel changes to the same part of the same file on two different branches, there will be trouble when they come back together.

Your local repository should know about the "remote" repositories namely your fork in GitHub, and Richard's original.  The first is probably called `origin` and the latter, if it's there is probably called `upstream` or `richardwilkes`, if not it would be a good idea to add it as a "remote" under one of those names.

I could tell you the specific command line instructions but I doubt it will help.  I don't know the tool you are using well enough to give detailed instructions so hopefully you can apply my description to what you have in front of you.

ethorad:
Awesome, many thanks

PR #209, opened 2023-01-21
Example npcs

ethorad:
Only some of the example NPCs in the back of the Basic Set were included in the library.  I've added the rest of them.


richardwilkes:
For William Headley, probably should remove the prereqs for Mental Strength, since the example clearly taking the "GM option" to approve for use by mages for this case.

ethorad:
Made that change now, thanks

richardwilkes (approved):

PR #210, opened 2023-01-27
Add Warhead types to Ultratech Equipment Modifiers

Legendsmith:
Both CF and basic set multiplication versions.


richardwilkes:
These edits look like they were done with an old version of GCS, since they have a key, “based_on_hash”, which I don’t believe is present any longer. Unfortunately, I’m currently on a business trip and can’t easily check this right now.

richardwilkes:
These changes look reasonable, except for the old format being used. Could you load them into the current version, save it out, then commit the result for this PR?

Legendsmith:
Thanks for the reminder, I've done that now.

richardwilkes (approved):

PR #213, opened 2023-01-27
Hogwarts characters

dripton:
From 1shotadventures.com's excellent free adventure "Harry Potter and the Warlock's Tunnel"


richardwilkes:
Perhaps this and the Prime Directive PR you recently put up should be organized under a `1shotadventures` folder to make it clear where they came from?

dripton:
On the one hand, if someone wants to run one of those adventures then having them under 1shotadventures makes it easier to find the characters.  OTOH, if someone is just looking for a sample character for a particular genre of game, then having them in a Characters folder under the genre is probably best.  Not really sure, so I'm happy to reorganize if you have a strong preference.

richardwilkes:
The current library is organized by book at the moment, not by genre, so I think I'd prefer something like `1shotadventures/Prime Directive/` and `1shotadventures/Hogwarts/` (or maybe `1shotadventures/Harry Potter and the Warlock's Tunnel/`, if that's the name of the adventure they came from).

Ideally, things like symlinks would be used to allow them to show up in both a book/adventure-specific location as well as a genre-specific one, but that's not really practical since not all platforms support them.

PR #222, opened 2023-02-28
Named Possession and Weapon Bond Perks from Fantasy

Lothex91:


Library/Fantasy/Fantasy Advantages.adq
```diff
@@ -19,6 +19,35 @@
 				"points": 2
 			}
 		},
+		{
+			"id": "379c5b53-3fa5-4728-8a3d-be657650cea6",
+			"type": "trait",
+			"name": "Named Possession",
+			"reference": "F132",
+			"tags": [
+				"Perk",
```
richardwilkes, line 28: Perks are defined to be 1 point advantages, just like Quirks are -1 point disadvantages. This tag doesn't seem appropriate here.

Library/Fantasy/Fantasy Advantages.adq
```diff
@@ -19,6 +19,35 @@
...
+				"points": 5
+			}
+		},
+		{
+			"id": "d3c15c2e-9bcd-4b9c-83c5-e75444a97a50",
+			"type": "trait",
+			"name": "Weapon Bond",
+			"reference": "F132",
+			"tags": [
+				"Perk",
```
richardwilkes, line 43: Here, either.

Lothex91: Sorry, my bad, I'm new to GURPS and for some reason kept thinking perks and quirks were 5 points advantages and disadvantages.

They are indeed perks, but I should correct their cost in 1 point.
Will do it right now and push the changes!

richardwilkes (approved):

PR #226, opened 2023-05-17
fixed weight and class errors

ZakariyaRguibi:
changed legality class of Trauma Plates for Reflex Tactical Vest from '42' to 2 [UT173]

and wight of Nanoweave Gloves from '6lbs' to negligible '0lbs' [UT172]


richardwilkes:
I haven't looked to see exactly why, but your commit has a very large number of changes. I suspect because you inadvertently changed all of the line endings from LF to something else, like CRLF. This needs to be fixed before I can accept the PR.

ZakariyaRguibi:
oh damn, cus it was a quick change so I changed it fro the UI. will fix it and resubmit shortly

PR #227, opened 2023-05-17
fixed weight and LCs errors

ZakariyaRguibi:
changed the legality class of 'Trauma Plates for Reflex Tactical Vest' from '42' to '2' [UT173]

removed weight of 'Nanoweave Gloves' to represent negligible '0lbs'  instead of '6lbs' [UT172]

### DIFF
--- a/Library/Ultra Tech/Ultra Tech Equipment.eqp
+++ b/Library/Ultra Tech/Ultra Tech Equipment.eqp
@@ -49524,7 +49524,7 @@
 			"description": "Trauma Plates for Reflex Tactical Vest",
 			"reference": "UT173",
 			"tech_level": "9",
-			"legality_class": "4",
+			"legality_class": "2",
 			"tags": [
 				"Armor",
 				"Defenses"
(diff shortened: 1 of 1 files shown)


Library/Ultra Tech/Ultra Tech Equipment.eqp
```diff
@@ -49525,7 +49524,7 @@
 			"description": "Trauma Plates for Reflex Tactical Vest",
 			"reference": "UT173",
 			"tech_level": "9",
-			"legality_class": "42",
+			"legality_class": "4",
```
richardwilkes, line 49527: Hmm... looking on page 173 of my copy of Ultra Tech, it lists all Tactical Vests and their trauma plates as LC 2, not 4. Do I just have an outdated PDF?

ZakariyaRguibi: oh, that's actually my bad one more time... I meant to write 2 instead of 4 in the description I say 2 but I wrote 5 in the code, is it okay for you to change it or should I revert and rechange and push?

richardwilkes: No need to revert -- you just need to edit it again and push the update.

richardwilkes (approved):

PR #230, opened 2023-06-08
Added missing modifiers from  PU8:5

Legendsmith:


Library/Power Ups/Power Ups 8 Limitations Limitation Modifiers.adm
```diff
@@ -40,6 +40,158 @@
...
+			"reference": "PU8:5",
+			"notes": "1d minutes of quiet chanting and simple materials",
+			"cost": -5
+		},
+		{
+			"id": "efaeec94-175b-4051-b659-5d0343bd5f0c",
+			"type": "modifier",
+			"name": "Requires Typical Ritual",
+			"reference": "PU8:5",
+			"notes": "1dx5\rminutes and @cumbersome preparation materials or successful casting of relevant spell (specify spell)@",
```
richardwilkes, line 85: This should not have a '\r' in it. GCS uses only line feed '\n' for embedded returns. If this was generated by GCS, it's a bug. Are you running on Windows?

Legendsmith: I am running windows, yes.
This was generated by GCS as well, yes. Very odd. This is what it appears like in GCS.
image

richardwilkes: Interesting. Did you copy & paste that from elsewhere, perhaps? I bet I don't filter out invisible characters and I probably should.

richardwilkes (approved):

PR #251, opened 2023-10-29
Add Centaurs from Dungeon Fantasy RPG Companion 2 and fix Large Fire Dragon filename typo

fishface60:
This adds Centaurs from [Dungeon Fantasy RPG Companion 2][].

I have speculatively used the page reference prefix `DFRC2:` because there is not currently an entry for it in the [GCS page references][].

The Striking ST for lower body that represents the horse part having much greater strength is not currently supportable as it is extra strength for Kicking and Slam attacks and strength for striking only isn't sufficiently granular.
In its place is a note to adjust kicking and slam damage manually.

In the Centaur monster sheet I have added extra damage to the hoof kick. I can't give it exactly the same damage because it would require an extra 1d+2 but I can't add extra dice of damage. +6 damage at least provides the same average and would result in the same if "Use Modifying Dice + Adds" is set.

The typo fix is included because I noticed it while looking through the monster character sheets, I can split it up into a separate PR if you would prefer.

[Dungeon Fantasy RPG Companion 2]: https://warehouse23.com/products/dungeon-fantasy-companion-2
[GCS page references]: https://gurpscharactersheet.com/page_references.html


richardwilkes:
You actually can add dice of damage. The issue is having a leading + causes it to think its a flat modifier, so instead, just put "1d+2" in the Damage Modifier field (i.e. not "+1d+2").

richardwilkes:
I've updated the website to include DFRPG Companion 2 with your suggested prefix.

fishface60:
> You actually can add dice of damage. The issue is having a leading + causes it to think its a flat modifier, so instead, just put "1d+2" in the Damage Modifier field (i.e. not "+1d+2").

Thanks for the tip. It turns out I was using the feature for adding a damage bonus to the weapon rather than modifying the attack to increase the base damage.

PR #252, opened 2023-10-30
Add Dungeon Fantasy 6 - 40 Artifacts Magic Armour

fishface60:
Most items in [Dungeon Fantasy 6 - 40 Artifacts][] do not have a listed price
and require a subjective process to asses their value,
but the Bracers of Force and Hooded Robe of Protection _do_ have listed prices and are useful for equipping spellcasters.

[Dungeon Fantasy 6 - 40 Artifacts]: https://warehouse23.com/products/gurps-dungeon-fantasy-6-40-artifacts

### DIFF
--- a/Library/Dungeon Fantasy/Dungeon Fantasy 6/Dungeon Fantasy 6 Equipment.eqp
+++ b/Library/Dungeon Fantasy/Dungeon Fantasy 6/Dungeon Fantasy 6 Equipment.eqp
@@ -1,0 +1,234 @@
+{
+	"type": "equipment_list",
+	"version": 4,
+	"rows": [
+		{
+			"id": "6a106f3c-5bbc-4c5a-b7b4-9e0f913c370c",
+			"type": "equipment",
+			"description": "Bracers of Force",
+			"reference": "DF6:5",
+			"reference_highlight": "Bracers of Force",
+			"notes": "11 FP as power item",
+			"legality_class": "4",
+			"tags": [
+				"Armor"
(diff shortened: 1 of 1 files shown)


Library/Dungeon Fantasy/Dungeon Fantasy 6/Dungeon Fantasy 6 Equipment.eqp
```diff
@@ -0,0 +1,234 @@
...
+	"type": "equipment_list",
+	"version": 4,
+	"rows": [
+		{
+			"id": "6a106f3c-5bbc-4c5a-b7b4-9e0f913c370c",
+			"type": "equipment",
+			"description": "Bracers of Force",
+			"reference": "DF6:5",
+			"reference_highlight": "Bracers of Force",
+			"notes": "11 FP as power item",
```
richardwilkes, line 11: It looks like most of the Bracers of Force have the same FP rating as a power item... shouldn't that vary depending on value?

fishface60: I saw that too. The Forcefield DR is some kind of enchantment and as I understand it enchantments don't count towards power item FP, and the bracers have a significant amount of DR themselves, so perhaps the material value accounts for the 11 FP.

At least, that's how I'm interpreting the section describing different forcefield DR values at different prices but only one listing for its worth as a power item.

PR #253, opened 2023-10-30
Add Dungeon Fantasy RPG Water Elemental

fishface60:
The full breakdown of how the attacks are built isn't provided, but I've inferred Drown is a feature of being grappled by a diffuse creature, Water Blast is a Double Knockback No Wounding Jet innate attack with reduced range, and Watery Blow is just a punch.

### DIFF
--- a/Library/Dungeon Fantasy RPG/Monsters/Elemental, Water.gcs
+++ b/Library/Dungeon Fantasy RPG/Monsters/Elemental, Water.gcs
@@ -1,0 +1,1283 @@
+{
+	"type": "character",
+	"version": 4,
+	"id": "69dfd924-5400-4fc6-9438-2ca41e351a0b",
+	"total_points": 150,
+	"points_record": [
+		{
+			"when": "2023-10-30T13:50:47Z",
+			"points": 150,
+			"reason": "Initial points"
+		}
+	],
+	"profile": {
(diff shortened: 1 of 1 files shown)


Library/Dungeon Fantasy RPG/Monsters/Elemental, Water.gcs
```diff
@@ -0,0 +1,1283 @@
...
+						"explanation": "Roll vs. Will to do anything besides talk or rest; failure causes unconsciousness\nEach FP you lose below 0 also causes 1 HP of injury\nMove, Dodge and ST are halved (B426)",
+						"ops": [
+							"halve_move",
+							"halve_dodge",
+							"halve_st"
+						]
+					},
+					{
+						"state": "Tired",
+						"expression": "round($fp/3)",
```
richardwilkes, line 163: I just noticed the FP & HP calculations for Tired and Reeling are using the much older, incorrect, formulas. You've likely not reset your "Default Attributes" to default since the fix was made. The same reset would also be needed on all sheets you've created... including the ones I just merged for you earlier this morning. :-(

fishface60: Thanks, I'll submit another PR to fix those.

PR #255, opened 2023-11-13
Add Scholar from Dungeon Fantasy 4 - Sages

fishface60:
I'm not sure what the best practice is for transcribing template options for "Raise other skill" or "worsen disadvantage". I've seen both listing the higher level and free-form text suggesting it.

I've happened to use both based on what felt most appropriate at the time, since it's fiddly to worsen a disadvantage, so easier to delete the less severe version, but it's trivial to raise a skill another level.

### DIFF
--- a/Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
+++ b/Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
@@ -1,0 +1,5861 @@
+{
+	"type": "template",
+	"version": 4,
+	"id": "e520671b-d0e0-4541-8d60-243372a4b345",
+	"traits": [
+		{
+			"id": "52691ea9-2893-4ba4-b85d-9a436522a9fb",
+			"type": "trait_container",
+			"open": true,
+			"children": [
+				{
+					"id": "580c0b70-b662-4816-a0a8-d0a5d8f1d10d",
+					"type": "trait_container",
+					"open": true,
+					"children": [
+						{
+							"id": "071894a8-8b79-447e-b69b-d8a3029e025d",
+							"type": "trait",
+							"name": "Increased Dexterity",
+							"reference": "B15",
+							"tags": [
+								"Advantage",
+								"Attribute",
+								"Physical"
+							],
+							"modifiers": [
+								{
+									"id": "26b39aca-fc45-404a-9dbd-be0e55323b67",
+									"type": "modifier",
+									"name": "No Fine Manipulators",
+									"cost": -40,
+									"disabled": true
+								}
+							],
+							"levels": 2,
+							"points_per_level": 20,
+							"features": [
+								{
+									"type": "attribute_bonus",
+									"attribute": "dx",
+									"amount": 1,
+									"per_level": true
+								}
+							],
+							"can_level": true,
+							"calc": {
+								"points": 40
+							}
+						},
+						{
+							"id": "b864937b-5a11-4a24-b7f9-ac55368094ef",
+							"type": "trait",
+							"name": "Increased Intelligence",
+							"reference": "B15",
+							"tags": [
+								"Advantage",
+								"Attribute",
+								"Mental"
+							],
+							"levels": 5,
+							"points_per_level": 20,
+							"features": [
+								{
+									"type": "attribute_bonus",
+									"attribute": "iq",
+									"amount": 1,
+									"per_level": true
+								}
+							],
+							"can_level": true,
+							"calc": {
+								"points": 100
+							}
+						},
+						{
+							"id": "71dbda15-cd71-45f1-824f-add678c6b151",
+							"type": "trait",
+							"name": "Increased Health",
+							"reference": "B14",
+							"tags": [
+								"Advantage",
+								"Attribute",
+								"Physical"
+							],
+							"levels": 2,
+							"points_per_level": 10,
+							"features": [
+								{
+									"type": "attribute_bonus",
+									"attribute": "ht",
+									"amount": 1,
+									"per_level": true
+								}
+							],
+							"can_level": true,
+							"calc": {
+								"points": 20
+							}
+						}
+					],
+					"name": "Attributes",
+					"container_type": "attributes",
+					"calc": {
+						"points": 160
+					}
+				},
+				{
+					"id": "72ae73c9-2d26-432c-9838-fd9541d84ab9",
+					"type": "trait_container",
+					"open": true,
+					"children": [
+						{
+							"id": "22cee029-fbfa-4e49-bd46-18b4a12494de",
+							"type": "trait",
+							"name": "Modular Abilities (Book-Learned Wisdom)",
+							"reference": "DF4:8,B71,P62",
+							"tags": [
+								"Advantage",
+								"Exotic",
+								"Mental"
+							],
+							"modifiers": [
+								{
+									"id": "ec37c7fa-ae90-4988-88c2-7c5649946fa4",
+									"type": "modifier",
+									"name": "Spells Only",
+									"reference": "B71",
+									"cost": -20,
+									"disabled": true
+								},
+								{
+									"id": "910a3a99-f36c-43bb-ba70-ba04df414dc2",
+									"type": "modifier",
+									"name": "Physical",
+									"reference": "B71",
+									"cost": 100,
+									"disabled": true
+								},
+								{
+									"id": "cc306036-66cf-4e12-ab17-423b26370304",
+									"type": "modifier",
+									"name": "Physical Only",
+									"reference": "B71",
+									"cost": 100,
+									"disabled": true
+								},
+								{
+									"id": "fb15a03b-7600-4294-9820-c4789280a88e",
+									"type": "modifier",
+									"name": "Virtual",
+									"reference": "B71",
+									"notes": "@Realm@",
+									"cost": -50,
+									"disabled": true
+								},
+								{
+									"id": "dca2e513-ac23-44cc-bb64-124382c743d1",
+									"type": "modifier",
+									"name": "Social Only",
+									"reference": "P63",
+									"disabled": true
+								},
+								{
+									"id": "26156d1d-ab8f-4e9b-93c3-1bf64c85e9af",
+									"type": "modifier",
+									"name": "Social",
+									"reference": "P63",
+									"cost": 50,
+									"disabled": true
+								},
+								{
+									"id": "763ea500-24dd-4b3b-b653-fb12b6069301",
+									"type": "modifier",
+									"name": "Focus-Limited (@Very versatile focus@)",
+									"reference": "P63",
+									"cost": -5,
+									"disabled": true
+								},
+								{
+									"id": "52e92e78-46bd-487d-8baa-539ad9419b6e",
+									"type": "modifier",
+									"name": "Focus-Limited (@Any ordinary variety of matter or energy@)",
+									"reference": "P63",
+									"cost": -10,
+									"disabled": true
+								},
+								{
+									"id": "b169a37a-0afa-41f2-8812-ee5c3f98a85f",
+									"type": "modifier",
+									"name": "Focus-Limited (@Specialized focus@)",
+									"reference": "P63",
+									"cost": -15,
+									"disabled": true
+								},
+								{
+									"id": "6d41662c-e079-4c9f-bf8a-4dec028b7425",
+									"type": "modifier",
+									"name": "Trait-Limited (@Subset of traits@)",
+									"reference": "P64",
+									"notes": "Set cost manually",
+									"disabled": true
+								}
+							],
+							"base_points": 5,
+							"levels": 2,
+							"points_per_level": 2,
+							"can_level": true,
+							"calc": {
+								"points": 9
+							}
+						},
+						{
+							"id": "7363031f-4453-422d-8f09-bcf6907d1cb2",
+							"type": "trait",
+							"name": "Modular Abilities (Book-Learned Wisdom)",
+							"reference": "DF4:8,B71,P62",
+							"tags": [
+								"Advantage",
+								"Exotic",
+								"Mental"
+							],
+							"modifiers": [
+								{
+									"id": "ec37c7fa-ae90-4988-88c2-7c5649946fa4",
+									"type": "modifier",
+									"name": "Spells Only",
+									"reference": "B71",
+									"cost": -20,
+									"disabled": true
+								},
+								{
+									"id": "910a3a99-f36c-43bb-ba70-ba04df414dc2",
+									"type": "modifier",
+									"name": "Physical",
+									"reference": "B71",
+									"cost": 100,
+									"disabled": true
+								},
+								{
+									"id": "cc306036-66cf-4e12-ab17-423b26370304",
+									"type": "modifier",
+									"name": "Physical Only",
+									"reference": "B71",
+									"cost": 100,
+									"disabled": true
+								},
+								{
+									"id": "fb15a03b-7600-4294-9820-c4789280a88e",
+									"type": "modifier",
+									"name": "Virtual",
+									"reference": "B71",
+									"notes": "@Realm@",
+									"cost": -50,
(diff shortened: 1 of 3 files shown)


Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
rinickolous, line 1: Please add attribute container type to "Attribute" trait container.

Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
rinickolous, line 1: You have several trait containers here with template choice fields which will not work when applying the template. While I understand that you are trying to communicate to the user that some traits are mutually exclusive, this should instead be done using prerequisites or not at all.

rinickolous, line 1: For skill containers where the user is given a choice between choosing a skill within the container or imporoving an existing skill, the quantity comparitor should be set to "is anything". Otherwise, the user must select one of the options even if they instead want to improve a skill. The name including "pick X amount or Y" should be sufficient indicaiton of what the user's choices are.

rinickolous (changes requested):
Good submission but requires some revision. Please see below:

Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
```diff
@@ -0,0 +1,4221 @@
+{
+	"type": "template",
+	"version": 4,
+	"id": "e520671b-d0e0-4541-8d60-243372a4b345",
+	"traits": [
+		{
+			"id": "52691ea9-2893-4ba4-b85d-9a436522a9fb",
```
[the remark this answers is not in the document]
fishface60: I had accidentally put them in the tags :sweat_smile:

Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
fishface60: Makes sense, I was adding them in explicitly because DF4 explicitly said to use one or the other, but I suppose they have to do that in the PDFs because people might not spot that.

fishface60: Makes sense. It would be nice to have a way to represent this without free-form text, since "Up to this level", "Replacing another trait", "upgrading a previous skill" and "more picks form a previous section" are pretty common, but the Book-Learned Wisdom would probably still be free-form and it doesn't sound simple to implement.

fishface60:
Thank you for the help @rinickolous

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
```diff
@@ -1245,8 +1244,9 @@
 					],
 					"calc": {
 						"level": 12,
+						"damage": "1d-3 cr",
 						"parry": "10",
-						"damage": "1d-3 cr"
+						"block": "10"
 					}
```
ethorad, line 1250: Punch should not have a block score

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
```diff
@@ -1281,8 +1281,8 @@
 					],
 					"calc": {
 						"level": 10,
-						"parry": "No",
-						"damage": "1d-2 cr"
+						"damage": "1d-2 cr",
+						"block": "9"
 					}
```
ethorad, line 1286: Kick shouldn't have a block score

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
```diff
@@ -2284,8 +2284,8 @@
 					],
 					"calc": {
 						"level": 13,
-						"parry": "No",
-						"damage": "1d+1 burn/point"
+						"damage": "1d+1 burn/point",
+						"block": "10"
 					}
```
ethorad, line 2289: A whole lot of your spell attacks have block scores and I don't think they should have

ethorad (changes requested):

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
```diff
@@ -1210,8 +1212,6 @@
 					],
 					"calc": {
 						"level": 12,
-						"parry": "No",
-						"block": "No",
 						"damage": "1d-3 cr"
```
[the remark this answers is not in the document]
fishface60: I think it's a result of the parry and block fields being validated now and previous conventions of entering no rather than leaving it empty.

Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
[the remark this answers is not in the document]
fishface60: > For some where there's no explicit limit, such as Acute Senses, I tend to put in a few levels only.

I'm not sure I understand the logic for this, can you confirm it is:

1. When levelled advantages are listed without an explicit limit in a block of "up to N points" that means you can buy up to any limit.
2. `Acute Senses (any) [2/level]` means any number of Acute Sense advantages at any level.
3. Because that can go up to 15 enumerating the possible values is impractical.
4. By relaxing the point limit to "Up to 30 points" the user can make selections and then fix the level later, but for convenience we enumerate up to a reasonable level so the user can pick from the list.

I'm not sure if a cluttered list is better for the user than learning to expect that levelled traits can be raised up to the limit, but since Perception is limited at +5 I've added that for the Acute Senses, and used the Will limit for Fearlessness.

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
fishface60: I think this sheet was created with an older version of the Natural Attacks. I've deleted it and re-added.

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
fishface60: I think this sheet was created with an older version of the Natural Attacks. I've deleted it and re-added.

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
fishface60: I tried refreshing the Wizard-Scholar's spell list from the list in Wizardry Redefined and it's the same from there, so wasn't a change on this character sheet or an outdated entry.

It seems to be that previously not having the block entry meant `"block": "No"` but now it means `"block": "0"`. Parry seems to be the same.

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
```diff
@@ -1225,7 +1225,6 @@
 					},
 					"usage": "Punch",
 					"reach": "C",
-					"parry": "0",
 					"defaults": [
```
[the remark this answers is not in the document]
fishface60: This line was removed because GCS now treats parry 0 as the default,

Library/Dungeon Fantasy/Dungeon Fantasy Adventure 1/Rivals/Mage - Wizard-Scholar.gcs
fishface60: This was because of GCS' change to make parry/block default to 0 instead of no. Re-adding Natural Attacks fixes it because it's not mis-loading from disk.

fishface60:
Thanks for the review @ethorad . It looks like the erroneous blocks are a result of the format incompatibly changing by making parry/block default to 0 instead of No.

I think most of the master library's data still expects the default of No instead of the default of 0, so this may be a bug gcs needs to fix first, or this change is going to result in any attack with block/parry 0 turning into No if it's loaded after the fix.

richardwilkes:
Hmm... previously, there were effectively 3 display states:

- The word "no"
- A number
- An empty field

That last state is no longer possible and the display always shows a number or the word "no". When saved to the file, the key/value pair will be omitted entirely if the bonus is +0.

Unfortunately, that collides with the previously allowed state of an empty field.

I could change this to treat an empty field as "no" instead. That would cause problems with newly-saved data, but would fix old data that should have said "no" explicitly.

I guess it really depends on just how many cases of that there are.

Library/Dungeon Fantasy/Dungeon Fantasy 4/Scholar.gct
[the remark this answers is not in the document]
ethorad: Agree that a cluttered list isn't great.  Especially where a template has "Acute Senses (any) [2/level]" with no real maximum.

ethorad:
For this template, it's probably easiest just to go through and click "No" on the relevant Block fields.

Hopefully when people use their files, seeing Block on a weapon is a flag to them that they need to update it.  Will leave it to @richardwilkes for whether this needs wider change.

fishface60:
> I guess it really depends on just how many cases of that there are.

In the Master Library, Basic Set Equipment had no empty parry or block until after you updated Shields, which affected 58 entries.

```
$ git cat-file blob c554f71a83a86bd72c5ad9ea72674c9b4da1dc73:Library/Basic\ Set/Basic\ Set\ Equipment.eqp | jq '..|select(type=="object" and .type=="melee_weapon" and (.parry == null or .block==null))|.id' | wc -l
58
```

The Elves and Goblins fantasy folk are the only other recent additions that may contain weapons.

```
$ find Elves/ Goblins\ and\ Hobgoblins/ -type f -exec jq -r '..|select(type=="object" and .type=="melee_weapon" and (.parry == null or .block==null))|@text "\(input_filename): \(.id)"' {} +
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: 751f75ea-84f7-4e40-bbdd-bade32678be5
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: edabce24-3b45-4ab7-9489-96be2a9ff796
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: 56243181-4724-42d6-8b95-3362a92238f4
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: 2f2ad09d-48bb-47a9-a070-d3594dad716f
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: f6d493e5-370b-4246-a5af-44e7758d0a02
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: 75c61c74-a34c-4f1a-8ad9-f67f7f882b7e
Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp: 46b558f4-6c61-47c2-9dd3-f9efad611b8c
Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct: 280a74a6-6d25-4625-9ff3-0d300cb58cd4
Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct: 2505fc43-09ca-4b2d-80b1-f153d51cfd6b
Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct: 65af3619-0f83-4ab2-bb1c-533bceb2175b
Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct: 647495f7-cdf5-49be-8cd1-cead9a2a6051
Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct: f8e67ca4-0c39-44de-a785-26ca251a43cc
```

The equipment is accurate, but Library/Goblins and Hobgoblins/Profession Templates/Hobgoblin Witch Doctor.gct has an erroneous block in Create Fire, Shatter, Smoke, Weaken and Hail spells.

There are 1087 weapons in the Master Library outside of Basic Set and Goblin Equipment which have an omitted block or parry, 

```
$ find . '(' -path ./Fantasy\ Folk/Goblins\ and\ Hobgoblins -o -path ./Fantasy\ Folk/Elves -o -path ./Basic\ Set/Basic\ Set\ Equipment.eqp ')' -prune -o -type f -a '!' '(' -name '*.md' -o -name '*.png' -o -name '*.html' -o -name '*.txt' -o -name '*.xml' ')' -exec jq -r '..|select(type=="object" and .type=="melee_weapon" and (.parry == null or .block==null))|@text "\(input_filename): \(.id)"' {} + | wc -l
1087
```

It's easy enough to batch fix them either way, though without preserving key order until they're re-saved.

Making Basic Set and Goblin Equipment explicitly set their block and parry to 0 so they can be loaded with current, past or a potential new version of GCS that changes the default to "No" can be done with:
```bash
for f in Basic\ Set/Basic\ Set\ Equipment.eqp Fantasy\ Folk/Goblins\ and\ Hobgoblins/Equipment/Goblin\ Equipment.eqp; do
  jq --tab 'walk(if type=="object" and .type=="melee_weapon" and (.parry == null or .block==null) then {block: "0", parry: "0"} + . else . end)' "$f" >"$f.new" && mv "$f.new" "$f"
done
```

Fixing everything else in the Master Library to fill in the parry and block defaults with No and resave can be done with:
```bash
find . '(' -path ./Basic\ Set/Basic\ Set\ Equipment.eqp -o -path ./Fantasy\ Folk/Goblins\ and\ Hobgoblins/Equipment/Goblin\ Equipment.eqp ')' -prune -o -type f -a '!' '(' -name '*.md' -o -name '*.png' -o -name '*.html' -o -name '*.txt' -o -name '*.xml' ')' -print | while read -r; do
  basename="$(basename "$REPLY")"
  tmp="${REPLY}.$$.${basename##*.}" &&
  jq --tab 'walk(if type=="object" and .type=="melee_weapon" and (.parry == null or .block==null) then {block: "No", parry: "No"} + . else . end)' "$REPLY" >"$tmp" &&
  gcs --convert "$tmp" &&
  mv "$tmp" "$REPLY"
done
```

I would lean towards there's probably more character sheets out there that will be opened again that haven't yet been saved in the new format, and sheets that have been saved in the new format are likely to be sufficiently actively used that bugs will be quickly noticed and fixed.

richardwilkes:
Thanks for the detail examination, @fishface60. Very helpful.

I just pushed out a GCS v5.20.1 release to address this (mostly). It now always saves out the parry & block fields and treats it as "No" if it is missing. This allows all the old files to be loaded without issue... but does mean anything saved with GCS versions v5.19.0, v5.19.1, or v5.20.0 will need some tweaks if they ran afoul of this issue.

fishface60:
I've updated this PR to the v5.20.1 format. I'll take a look at fixing the equipment libraries if nobody has already beaten me to it.

fishface60:
I've updated this PR to the v5.20.2 format, which removes the `"parry": "No"` and `"block": "No"` from melee and ranged weapons.

fishface60:
I've rebased on top of master and corrected the Natural Attacks to that of v5.20.3 so the reach is correct, though I'm not familiar with unarmed having a block, so that might need correcting.

ethorad (approved):

rinickolous (approved):
Thanks for the changes. Approved.

PR #259, opened 2023-11-23
Add "Bardic" tag to Lend Language in DFRPG spells.

dripton:
This allows Bardic Talent to add to the spell level.


rinickolous (approved):
All good, thanks for the fix.

PR #261, opened 2023-12-12
Fix Low-Tech abdomen armor cost and weight

dec-00:
All abdomen armor pieces from Low-Tech had their cost and weight input as if they were 75% of that material's base value. It should be 25% according to Low-Tech, p. 100.


rinickolous (approved):
Looks good. Thank you for the contribution.

PR #262, opened 2023-12-16
Fix Easy Skill Modifier Value (PU8 Limitations)

Legendsmith:
-10% ``->`` -5%
Power Ups 8 Limitations Limitation Modifiers.adm


rinickolous (approved):
Approved.

PR #264, opened 2023-12-23
Add support for Low-Tech's chest and abdomen hit locations and armor

dec-00:
* Added the _Humanoid (Low-Tech)_ body type. This body type separates the _torso_ hit location into _chest_ and _abdomen_ as per GURPS Low-Tech. The _chest_ still has the ID of "torso", to maintain as much compatibility with non-Low-Tech armor pieces as possible, so any armor piece that adds DR to the default body type's _torso_ will add it to this body type's _chest_.
* Updated _Low Tech Armor (by location) and (by material)_ to support this new body type, while maintaining compatibility with the default body type.
    * Armor that covers the "torso" now adds DR to the _abdomen_ and _groin_ locations, as the groin is considered part of the abdomen in Low-Tech, hence also part of the torso.
    * Armor that covers the "chest" now adds DR to the _chest_ (id: torso) and _vitals_ locations.
    * Armor that covers the "abdomen" now adds DR to the _abdomen_ and _groin_ locations. Technically the vitals can be targeted through the abdomen as well as the chest, but I decided against adding vitals DR to abdomen armor, as it would stack with vitals DR from chest armor.
* Added DR to the _abdomen_ location for all torso armor in Low-Tech Instant Armor. They already correctly had groin DR.
* Added all chest and abdomen armor pieces from GURPS Low-Tech Instant Armor.

### DIFF
--- a/Library/Low Tech/Low Tech Armor (by location).eqp
+++ b/Library/Low Tech/Low Tech Armor (by location).eqp
@@ -2654,6 +2654,16 @@
 					"features": [
 						{
 							"type": "dr_bonus",
+							"location": "abdomen",
+							"amount": 1
+						},
+						{
+							"type": "dr_bonus",
+							"location": "groin",
+							"amount": 1
+						},
+						{
+							"type": "dr_bonus",
 							"location": "torso",
 							"amount": 1
 						},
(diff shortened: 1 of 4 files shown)


rinickolous (approved):
Looks good. I was initially tentative about the way you went about organising the hit locations, but I don't think there's a better way of going about it that won't break non-Low-Tech compatibility.
Approved.

PR #266, opened 2024-01-01
Add Fantasy Folk - Elves to the library

ethorad:
Added the traits, racial templates and profession templates from the FF: Elves book


rinickolous (changes requested):
As with your other submission, I would put this in a subfolder under "Fantasy Folk".
Miscellaneous notes:
- Attribute Substitution in "Elvish Traits" has incorrect text substitution format.
- I would remove "Elbish Typical Traits".

ethorad:
Happy to put in a subfolder.  Should I move the existing templates in the "Fantasy Folk" folder to a subfolder "3 Edition templates" to keep it clean?

Also do you have an example of what the text substitution format should look like?

ethorad:
Made the change for the file location, and removed the typical traits.

I've had a look around the library in some likely places (Powers, Power Ups, a couple of others) but I can't spot somewhere else with Alternative Attribute, so not sure what the right text substitution format is.

ethorad:
Actually, noticed I used % in the Alternative Attribute instead of @ for the text substitution - think that's what you were referring to.  I thought it was some way to get it to amend the Innate Attack skills to reference Per instead of DX which I wasn't aware of.

Made that change

rinickolous (approved):
Thanks for the amendments. Approved.

PR #267, opened 2024-01-01
Fantasy Folk - Goblins and Hobgoblins

ethorad:
Added the traits and templates from the Fantasy Folk - Goblins and Hobgoblins book


rinickolous (changes requested):
For tidiness sake, I would put this in a subfolder under "Fantasy Folks".
Miscelleaneous notes:
- Martial Art: Double Knife is missing a page reference
- For your Martial Arts templates, I would utilise the template choice fields. You're welcome not to do this, but it would be helpful.
- I would remove the file "Goblin Typical Traits". It seems unnecessary.
- War-Beast "Attributes" trait container is missing "Attributes" container type.

ethorad:
I didn't use the template choice groups for the Martial Art as I wanted to keep the traits and skills there for future reference.

With the choice approaach, you get one chance to select the traits and skills when you first apply the template.  They are then gone and if you want to develop your character later you need to go back to check what the style components are.

With them all in the sheet in a group but disabled, then it acts as a reminder that you can easily turn on.  If you want (e.g. if you know you're unlikely to take certain traits) then they can be deleted after applying the template.

I can convert them to use option groups though.  Would set to "Choose any" and have separate groups for the core traits/skills and then another option group for the traits/skills which are actually optional for the style so it's clear to the user which are more essential.

What's the preference?  I did see that the MA book doesn't seem to have any styles in there so I didn't have a template layout to copy

ethorad:
Made the changes (not sure if you got notified so thought I'd add a comment)

rinickolous (approved):
Your justification for the way Martial Arts styles are done makes sense, and since there's no real precedent on format I guess we'll just go with it.

Approved.

PR #268, opened 2024-01-01
Moved 3Ed Fantasy Folk files

ethorad:
As we are creating directories for the 4E Fantasy Folk books, I thought it best to move the 3Ed files to their own subdirectory


rinickolous (changes requested):
Should rename to "3rd Edition Races". "3 Edition Races" seems grammatically incorrect

ethorad:
Changes done, thanks

rinickolous (approved):

PR #269, opened 2024-01-03
Fix parry and block missing from v5.19.0-v5.20.0 edits

fishface60:
This mechanically fixes the missing Block and Parry scores for Basic Set and Goblins equipment.

```bash
for f in Basic\ Set/Basic\ Set\ Equipment.eqp Fantasy\ Folk/Goblins\ and\ Hobgoblins/Equipment/Goblin\ Equipment.eqp; do jq --tab 'walk(if (type=="object" and .type=="melee_weapon" and (.parry == null or .block==null)) then {block: "0", parry: "0"} + . else . end)' "$f" >"$f.new" && mv "$f.new" "$f" && gcs --convert "$f"; done
```

This is sufficient on its own, though `gcs --convert` currently also adds parry and block scores to ranged attacks, so I've filtered those changes out of the patch.

### DIFF
--- a/Library/Basic Set/Basic Set Equipment.eqp
+++ b/Library/Basic Set/Basic Set Equipment.eqp
@@ -1467,7 +1467,6 @@
 					"usage": "Swung",
 					"reach": "1",
 					"parry": "0U",
-					"block": "No",
 					"defaults": [
 						{
 							"type": "dx",
@@ -2184,7 +2183,6 @@
 					"usage_notes": "1-Handed",
 					"reach": "1-2",
 					"parry": "0U",
-					"block": "No",
 					"defaults": [
 						{
 							"type": "dx",
@@ -2241,7 +2239,6 @@
 					"usage_notes": "1-Handed",
 					"reach": "2",
 					"parry": "0U",
-					"block": "No",
 					"defaults": [
 						{
 							"type": "dx",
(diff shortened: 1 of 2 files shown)


richardwilkes:
Argh! I thought I checked that after making that last change. It won't hurt anything, but it is annoying.

richardwilkes:
... and it looks like the fix I made doesn't cover all the cases I thought it did for the old data. :-(

fishface60:
I've updated to the v5.20.2 format. Apart from now omitting the `"parry": "No"` this also removes the `"level"` and `"block"` fields from the `"calc"` part of cloaks, which appears to be fine.

Library/Basic Set/Basic Set Equipment.eqp
ethorad, line 1: Been through everything with tag "Melee Weapon" and confirmed it has the right Parry and Block stats.  Similarly for everything with tag "Shield".

Library/Fantasy Folk/Goblins and Hobgoblins/Equipment/Goblin Equipment.eqp
ethorad, line 1: All weapons have right Parry and Block stats

ethorad (approved):
Been through the basic set equipment tagged as Melee Weapon or as Shield.  Don't think there will be other stuff which has Parry and Block stats.  Happy with the Parry and Block entries on all of them.
Also happy with the Goblin equipment changes (although this may have been fixed in a different PR)

PR #272, opened 2024-01-04
Some changes to the Goblin files allow for latest GCS version

ethorad:
Equipment file:
Amended all weapons to have +0 parry (except for some of the Glaive ones which retained their +2 parry)

Goblin Emissary:
Removed a "Placeholder" trait I use to make it easier to add new traits to the end in the right group.

Hobgoblin Witch Doctor:
Didn't really make any changes as it all looked OK with no parry or block permitted on the spells.  Saving it seems to have removed those rows from the save file so it shows up as a change.


ethorad:
Think this might be be largely superfluous given the change I've just accepted from fish.  As such the only change is probably removing the place holder trait I left in one of the templates

rinickolous (approved):
Approved.

PR #275, opened 2024-01-12
1/12/2024 Update

incrsnentrpy:


Library/Discworld Roleplaying Game/Traits/Discworld Advantages.adq
rinickolous, line 1: Please make the following changes:
- Split "Vampire (Male)" and "Vampire (Female)" into separate files.
- Create attribute containers for attribute modifiers, with "Attributes" container type.
- Change page references to match those found in the DW sourcebook as opposed to Basic Set.

rinickolous (changes requested):
Pertaining to all of the character files: Please amend any page references to match those found in the DW sourcebook, considering that it is self-contained.

Other than that, please make the changes to individual files detailed below.

PR #276, opened 2024-01-13
Update Druid.gct

daliin:
DFRPG Druid template was lacking the two special skills, Esoteric Medicine (Druidic) and Herb Lore that appear at the end of the Druid template in DFRPG Adventurers (p. 23).

### DIFF
--- a/Library/Dungeon Fantasy RPG/Classes/Druid.gct
+++ b/Library/Dungeon Fantasy RPG/Classes/Druid.gct
@@ -115,6 +115,7 @@
 						}
 					],
 					"name": "Attributes",
+					"container_type": "attributes",
 					"calc": {
 						"points": 160
 					}
@@ -151,6 +152,7 @@
 						}
 					],
 					"name": "Secondary Characteristics",
+					"container_type": "attributes",
 					"calc": {
 						"points": -5
 					}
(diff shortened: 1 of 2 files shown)


rinickolous (changes requested):
Please change the container type of the 2 attribute containers to "Attributes". Other than that, no issues.

daliin:
Changed the two attribute containers to Attributes. Good catch.

daliin:
Hmm. Apparently I don't fully understand how pull requests work. I meant for the Eldhud racial template changes to be a separate thing, but they've apparently been lumped into the druid request.

rinickolous (changes requested):
Thanks for the changes. If you could just remove the container saying "Can purchase up to Fire Resistance 15" from the Eldhuhd template as it's non-standard, then this will be approved. While I understand your intention with the container, we don't have consistency across other files with this at the moment.

daliin:
> Thanks for the changes. If you could just remove the container saying "Can purchase up to Fire Resistance 15" from the Eldhuhd template as it's non-standard, then this will be approved. While I understand your intention with the container, we don't have consistency across other files with this at the moment.

Makes sense. I removed the container.

rinickolous (approved):

PR #278, opened 2024-01-15
DF Druid to Druidic

daliin:
Changed Power Investiture (Druid) to Power Investiture (Druidic) to match the terminology used in the rules and DFRPG. Adjusted it in the Traits and Spells libraries, Druid template, and the  rival druid in Mirror of the Fire Demon.


rinickolous (approved):
Sorry for the late review, was weired out by the "unsatisfied reason" bits added to the Druid's spells and didn't have time to actually open GCS and check it out.

Approved.

PR #279, opened 2024-01-17
Magic Spells : Add History prerequisite to Dinvination Spells.

Scimon:
In the Magic book page 108 Divination has a prequisite of : History, and other spells as specified for the particular method of divination.

Unfortunately the History prequisite was missed off in the library.


rinickolous (approved):
Change approved. Thanks for the fix.

PR #285, opened 2024-01-27
Wingedfolk

ethorad:
Adding Fantasy Folk: Winged Folk files


rinickolous (changes requested):
Your .gcs files include non-default attribute settings. I'm pretty sure these were settings adopted as defaults temporarily but reverted after some time. Probably best to change them to the current defaults.

ethorad:
Fixed that, thanks

rinickolous (approved):

PR #286, opened 2024-01-29
Transhumanupdate

ethorad:
removed the two offending files, and fixed a type in the name of the reporter's vest in the transhuman files

(note because I made some changes on my master fork rather than on a branch, these new changes are mixed in with some other ones I did - sorry!)


rinickolous (approved):

rinickolous:
> Out of curiosity, did the removed files had an actual source book or were they just made up?

They were apparently part of someone's homebrew game and added by mistake.

Issue #271, opened 2024-01-03
Prefect Balance should give its bonus to Aerobatics and Aquabatics as well as Acrobatics


Rupert01:
From B174:

> Acrobatics, Aerobatics, and Aquabatics default to one another at -4. Add +1 to these three skills if you have Perfect Balance (p. 74).

PR #288, opened 2024-02-07
Fix for perfect balance - Issue #271

Lumek4:
As per Basic Set - Characters p. 174 under "Acrobatics", you should add +1 to Aquabatics and Aerobatics as well.


ethorad (approved):
Happy with the change

ethorad:
Merged in, many thanks

PR #289, opened 2024-02-09
Add Mana Sensitive

Lumek4:
Limitation that comes from "Advantage Origins", "Magic", on pages 33/34 of Basic Set - Characters.


ethorad (changes requested):
Happy with the added limitation, however can you move it to be in alphabetical order?  So just after Limited Use and before Melee Attack.

(I would change, but I can't merge things I've changed)

Thanks!

ethorad (approved):
Many thanks for that, will merge that change into the library

PR #292, opened 2024-02-26
Update Evil Cleric.gct

Legendsmith:
- Changed the Cleric Spells prerequsites so they work with the Power Investiture on the sheet. It is "Power Investiture (Unholy)", but the spells *were* looking for exactly "Power Investiture", so I changed them to "starts with."
- Updated the Dread Touch trait to the correct cost, and weapon type. (Yes, it needs round down set to true, as its cost in the book is 14 but without rounding down it's 15)

### DIFF
--- a/Library/Dungeon Fantasy/Dungeon Fantasy 3/Evil Cleric.gct
+++ b/Library/Dungeon Fantasy/Dungeon Fantasy 3/Evil Cleric.gct
@@ -1963,6 +1963,8 @@
 													"id": "7ae9aa1e-141d-4963-8048-f6542618be93",
 													"type": "modifier",
 													"name": "Cosmic, Irresistable Attack",
+													"reference": "B103",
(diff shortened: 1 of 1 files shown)


Library/Dungeon Fantasy/Dungeon Fantasy 3/Evil Cleric.gct
ethorad, line 1: The power investiture 1, 2 and 3 spells need updating still.

ethorad (changes requested):
Thanks for this

An alternative approach could be to have the "Unholy" in the notes section of the Power Investiture.  Then the spells can look for a trait called "Power Investiture" where the notes section is "Unholy".  I think that would be cleaner, but probably not worth you redoing the changes you've already made, so fine to stick with looking for a trait that starts with "Power Investiture"

dripton:
Any followup on this one? Looks like it was left hanging.

PR #295, opened 2024-03-14
Add REC7 to and from TacShooting

fdenzer:
TS63 (table) TS64 (variant, ammo)

### DIFF
--- a/Library/Tactical Shooting/Tactical Shooting Equipment.eqp
+++ b/Library/Tactical Shooting/Tactical Shooting Equipment.eqp
@@ -2419,6 +2419,124 @@
 			}
 		},
 		{
+			"type": "equipment",
+			"description": "6.8x43mm Special Purpose Cartride (SPC)",
+			"reference": "TS64",
+			"tags": [
+				"Ammunition",
+				"Warheads"
+			],
+			"quantity": 1,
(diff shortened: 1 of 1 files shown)


rinickolous (commented):
This all looks fine, but would like to understand why the weight for both this and the other already present variant differ by 0.3lb from the book. I also checked a few other entries, which differ by different amounts. I assume this has something to do with WPS and magazine weight, but the values don't seem to add up correctly.

fdenzer:
Are we putting loaded or empty weight in the equipment table?
IIRC the rules give empty weapon weights, but "packs of ammo including their magazines" as A/B notation, like 7/1 for an assault carbine, starting in Basic. But WPS allows to calulate empty magazines weight, starting in high tech. And WPS times ammo count should add to empty rifle weights to get a loaded weight, which is, again IIRC, never specified in RAW.

richardwilkes:
Personally, I'd want the empty weight for the weapon, as well as entries for an empty magazine and the per-shot weight for ammo. Others may not want such detail, but they can always compute the aggregate and adjust, whereas going the opposite direction is much harder.

dripton:
Any followup on this one? Looks like it was left hanging.

PR #307, opened 2024-06-21
CDRU's home brew TL9 missile weapons

dripton:
gcs file by Anders, with permission from CDRU


richardwilkes (commented):
Could I get you to replace the underscores in the file name with spaces?

richardwilkes (approved):
I've not taken a look at the file yet as I'm swamped with work right now... but trust your judgement on its content.

PR #309, opened 2024-07-14
Update for the v5 data format used in GCS v5.24+

richardwilkes:


richardwilkes:
I'm omitting the work necessary to update all of the templates for now. Ideally, each existing template would have its contents replaced with data pulled from the relevant libraries so that the source library can be tracked. I don't have the time or energy for doing such a thing right now, but we can slowly update them over time.

Issue #312, opened 2024-07-27
Talents in the PU3 library all include BOTH the reaction bonus and the alternate benifit bonus


smithkm:
PU3 provides an alternative benefit for each talent to use instead of the reaction bonus. They used to be modifiers so they could be easily switched between.  It looks like they have been moved to being features on the base trait which makes both show up.  I expect this was done to allow them to include the level of the talent in their condition messages.

Either each talent needs to be included twice, once for reaction bonus and once for alternate benefit  (A gcs_master_library) or we need a way to toggle between different features on a trait like we can modifiers, or the condition message for a modifier needs to be able to use the level of its parent trait instead of its own level.  (Any of which would require a change upstream in GCS see https://github.com/richardwilkes/gcs/issues/660)

richardwilkes:
Does this still need some work (i.e. data file changes or code changes in GCS)?

PR #313, opened 2024-08-03
#312 Talent benefits as modifiers

smithkm:
Returned the alternate benefits of talents in the Power Ups library to being modifiers so they can be easily switched between.  For talents that have no alternate benefit I left  the reaction bonus directly on the trait for simplicity.

### DIFF
--- a/Library/Power Ups/Power Ups Traits.adq
+++ b/Library/Power Ups/Power Ups Traits.adq
@@ -7219,6 +7219,41 @@
 				"Mental",
 				"Talent"
 			],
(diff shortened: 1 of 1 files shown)


dripton:
I agree with the bug report.  The diff looks reasonable, though it's large enough that I can't confirm that there are no typos.  Looks worth merging.

dripton (approved):
I looked at the bug and the diff and this seems sensible.  It's big enough that I can't 100% confirm there are no typos, but we can always tweak if needed.

PR #324, opened 2024-10-27
Fix #247: Fix wrong book abbreviation for Transhuman Space

### LINKED ISSUE #247: Wrong book abbreviation for Transhuman Space
Hi,

this is my first reported bug, so please be gentle with me but feel free to let me now, if I should do something differently next time.

I noticed that in the Master Library (v4.16) several traits (e.g. all Infromorph templates) the page references use the abbreviation TS (Tactical Shooting), not THS, which it should be according to https://gurpscharactersheet.com/page_references.html

I would guess that a replace in the "reference" tags (TS->THS) should fix the problem ("TS" is not part of any other relevant references).

On a personal note: I've played GURPS since the 90s and always used the GCA, which seems to be defunct now. So I was very happy to find the GCS project after a hiatus. Thank you for the great work. I might try to help fill the Master Library out (maybe with more Transhuman Space stuff) if a get some free time.

Cheers,
    Nils.

richardwilkes:
Fix #247

### DIFF
--- a/Library/Transhuman Space/Bioroids/Algonaut.gct
+++ b/Library/Transhuman Space/Bioroids/Algonaut.gct
@@ -709,7 +709,7 @@
 				{
 					"id": "TC5sPaFSiiB2gNktR",
 					"name": "Bioroid Body",
-					"reference": "TS131",
+					"reference": "THS131",
 					"tags": [
 						"Bioroid",
 						"Meta-Trait"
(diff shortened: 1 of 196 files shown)


dripton:
I verified that TSH is the right abbreviation from https://gurpscharactersheet.com/page_references.html .  I did a git grep to find 399 lines with the wrong prefix, and saw that your patch has 399 lines of fixes, so it looks right.  I didn't review the actual page numbers though.

dripton (approved):
I verified that TSH is the right abbreviation from https://gurpscharactersheet.com/page_references.html . I did a git grep to find 399 lines with the wrong prefix, and saw that your patch has 399 lines of fixes, so it looks right. I didn't review the actual page numbers though.

dripton:
(I see there was a race where I approved your first commit while you added 2 more, but the others are much smaller and look correct.)

PR #325, opened 2024-11-02
Added Broken Empire data from Donald Qualls

richardwilkes:


dripton (approved):
It's a lot of skills and a lot of equipment.  Looks legit.  It's Home Brew so I won't worry about the exact content much.

PR #329, opened 2024-12-03
Ein arris pregens

derrickthewhite:


dripton (changes requested):
I don't see any of these characters in the 4E version of Caravan to Ein Arris.  What is the source?

derrickthewhite:
> I don't see any of these characters in the 4E version of Caravan to Ein Arris. What is the source?

They are from the worlds beyond earth blog (https://worldsbeyondearth.blogspot.com/). Back when I made them Rich Suggested that I put them in the Library. 

Here is where we talked about it (over a year ago): 

https://discord.com/channels/731926683371700315/1081628926675275918/1185303382710296576

dripton:
I think these should go under "Home Brew/Characters".  Putting them under "Basic Set" makes them look official.

derrickthewhite:
That's a great Idea. I don't think Homebrew was around when I Iooked at this a year ago. I'll move it there. Thanks

dripton (approved):
Now that these are moved to Home Brew, I loaded all the characters into a recent version of GCS.  All loaded fine and look good.  I think these would be a good addition to the master library.

dripton:
No comments from anyone else so I will try to merge this.

dripton:
Hmm, the "close" button didn't seem to merge it, so I reopened it.  Rich will see this and click whatever needs clicking.

Issue #314, opened 2024-08-06
Would these Aliens-Sparrials (4e) files be acceptable?


ArchangelBeth:
(I started crying during the youtube video, I'm sorry, I'm sorry. ;_; )
Fingers crossed that this is attaching the files okay.

[Aliens-Sparrials and Settings.zip](https://github.com/user-attachments/files/16516551/Aliens-Sparrials.and.Settings.zip)

richardwilkes:
OK, I'll take a look when I get some spare time.

richardwilkes:
I took a quick glance today and noticed that the page references are mostly referring to things not currently present in the [GCS Page Reference List](https://gurpscharactersheet.com/page_references.html) and the few that I could find that should have been using the `AS` prefix (for the 4th Edition GURPS Aliens: Sparrials), aren't.

The only other thing I had a chance to start going through was the Sparrial racial template, and I noticed it didn't add up to the value in the book. In particular, it lists Increased Consumption for -10 points in the book, but you have it at 0 -- looks like the appropriate modifier for level wasn't enabled (ideally, this would just be a standard leveled trait, but it looks like whoever entered it for the basic set data wanted to list the consumption it caused, which is derived from the level, but not just a +X value).

@dripton @rinickolous @ethorad : Do any of you have the time and desire to go over these and help @ArchangelBeth get them into the Master Library? If not, I'll eventually get around to it, but I'll admit to not being super-familiar with space stuff and I'd rather focus my time on more GCS development work.

richardwilkes:
Oh, and I should have noted, if the GCS Page Reference List is missing some 4th Edition books that we need to add a code for, I'm happy to pick something and add it. Just let me know.

ArchangelBeth:
•makes note of what I'm doing on Monday• Thanks!

ArchangelBeth:
Well, it's after _a_ Monday. Somewhere. Sorry it took so long; got roped into GMing for a game that required setup and then it fell under the table. I _think_ I have addressed the issues mentioned. If my eyes have betrayed me and I missed one... Apologies.
[Aliens-Sparrials and Ancestries.zip](https://github.com/user-attachments/files/18271875/Aliens-Sparrials.and.Ancestries.zip)

richardwilkes:
Merged the data. Will be part of the next library release.

PR #332, opened 2024-12-31
Sparrials

richardwilkes:


richardwilkes:
Added for @ArchangelBeth from issue #314. Did a bit of rearranging and some minor editing (removing her name from the Player field for some of the animals/npcs, for example).

dripton (approved):
I don't have the Sparrials book so I can't review in detail, but this all looks reasonable.  I'd let it in and see if anyone else finds problems.

PR #333, opened 2025-01-14
Add Magical Bolt advantage from Dungeon Fantasy 11

dripton:
Did this as 6 advantages because I didn't see an easy way to make this a leveled trait.  (The cost and damage go up linearly for the first 4 levels but then get non-linear.)


richardwilkes:
The magical bolt ranged weapon entry needs to have a default set for it (in this case, to the Innate Attack (Projectile) skill) so that it shows the effective skill level.

richardwilkes:
Should probably also note the info from the book about what a hit means (i.e. ignores DR, does harm golems, spirits, undead, etc., while completely inanimate targets are unaffected)

dripton:
Yeah, I'll add the skill entry.  I'll also add a note that it ignores DR.

richardwilkes (approved):

PR #337, opened 2025-01-19
Update Dragon-Blooded.gct

daliin:
Update Dragon's Breath Advantage to include the 12 point cost. Also added a bit of descriptive text.


dripton (approved):
This looks correct.

PR #338, opened 2025-01-24
Add stringed instruments to DFRPG Equipment

daliin:
Added the cittern, lute, and oud to DFRPG Equipment from Adventurers, p. 112.


dripton (approved):
Looks good

PR #339, opened 2025-01-25
Create Alphabet Arcane Spells.spl

daliin:
Added the five new spells from Alphabet Arcane.


dripton (approved):
I don't have Alphabet Arcane so I can't check these all for accuracy, but they look well-formed.

PR #342, opened 2025-02-02
Improve prerequisites for Gift of Letters and Gift of Tongues

dripton:
The requirement in GURPS Magic is Borrow Language plus three languages at Accented or better (written for Letters, spoken for Tongues).  Note that GCS shows languages like "Language: English" so the right way to check for a language is "starts with: 'Language:'". (If you miss the colon then the Language Talent advantage is a false positive.)

Previously the GCS Magic version of these spells just required one language at Accented or better, which was wrong.  I've changed it to three languages and comprehension other than None, which is closer but still not quite correct since Broken comprehension is not supposed to count.  I don't see a way to express "3 or more languages at Accented or better" within the current GCS system; it would require nested levels of boolean logic, or mapping the language levels to numbers.

Dungeon Fantasy RPG doesn't have Broken or Accented languages, so "3 or more languages with Spoken/Written not set to None" is correct. (And the alternate clerical prerequisite of Power Investiture 4 is kept.)

### DIFF
--- a/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Spells.spl
+++ b/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Spells.spl
@@ -6071,30 +6071,6 @@
 												}
 											},
 											{
-												"type": "trait_prereq",
-												"has": true,
-												"name": {
-													"compare": "starts_with",
-													"qualifier": "Language:"
-												},
-												"notes": {
-													"compare": "does_not_contain",
-													"qualifier": "Written (None)"
-												}
-											},
-											{
-												"type": "trait_prereq",
-												"has": true,
-												"name": {
-													"compare": "starts_with",
-													"qualifier": "Language:"
-												},
-												"notes": {
-													"compare": "does_not_contain",
-													"qualifier": "Written (None)"
-												}
-											},
-											{
 												"type": "prereq_list",
 												"all": false,
 												"prereqs": [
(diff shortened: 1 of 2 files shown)


richardwilkes:
I don't believe there is any support for what you're trying to do (find N copies of a trait with some set of properties). There is support for that type of thing with spells for certain subsets of criteria. Just listing it multiple times won't have any effect different from listing it once.

As for nested levels of boolean logic, that is supported.

Anyway, bottom line, I don't believe this change actually does what you think it does. Have you tested the result?

dripton:
The "3 copies of Language" logic was already there in the DFRPG version.  After more testing, I agree it doesn't actually work as intended.

The other fix, using "starts with 'Language:'" instead of "is Language", works correctly and fixes a bug that caused the DFRPG version of these spells to never work for wizards/bards.

I didn't see how to nest multiple levels of boolean logic, but if you say it's there, great, I just didn't figure out the GUI. But without the ability to count Languages, it doesn't really help for this case.

I will simplify this PR to only include the part that actually helps and not include the part that changes things without actually helping.

dripton:
I made a second commit to remove the change to add 3 copies of "Language:" to the GURPS Magic version (because it doesn't work), but keep the "starts with Language:" (to ensure that Language Talent isn't treated as a language).

richardwilkes:
Looks better, although I'd also remove the extra copies that don't do anything (i.e. just one of each, instead of 3).

richardwilkes (approved):

PR #343, opened 2025-02-02
Create Hand of Asgard Traits.adq

daliin:
Added the Holy Might abilities from Hand of Asgard.


dripton (approved):
I don't have this adventure so I can't fully check these for correctness, but they look right and I didn't spot any typos.

dripton:
I'll wait a bit to see if we get other reviews before merging.

PR #347, opened 2025-02-11
Add various human ancestries to Arden Vul home brew.

dripton:
Since we don't use Cultural Familiarities in DFRPG, this really only affects native languages.


richardwilkes:
These look fine to me, although you could add some custom .ancestry files so that the various parameters in your notes (height ranges, coloration, etc) are respected by the randomizer.

richardwilkes:
I'll go ahead and approve in case you don't want to bother with that extra work, though.

dripton:
Custom .ancestry files are a great idea, though I think I'll merge this much first then come back to them.

PR #350, opened 2025-03-10
Add Chi Rations

dripton:
DFA 30 Chi Abilties says people with them need to pay double for rations.  So add Chi Rations to the DFRPG equipment list.  They're just Rations with the price doubled.  These will keep Chi Abilties functioning.  Other characters can eat them too; they just provide no special benefits to them.


richardwilkes:
Is this because they are made of comparatively exotic food, or is it just a need to consume twice as much? If the latter, I would expect the weight to double as well.

dripton:
It's health food (picky about ingredients and quality), not more food.  So twice the cost for the same weight.

richardwilkes (approved):

PR #354, opened 2025-03-27
Ultra-Tech: Make swarmbots more customisable without breaking sync with the source library

maw3193:
This covers broadly two things:
1. Making it possible to customise swarmbot chassis' and power supply without editing the equipment (which would cause problems syncing with the source library)
  * This adds equipment modifiers for swarmbot chassis', including cost multipliers and notes on how this affects their movement.
  * This adds equipment modifiers for swarmbot power supplies, including cost multipliers and notes on how this affects their endurance.
3. Making it possible to specify what items repair swarmbots can repair without having to edit the equipment (which would cause problems syncing with the source library)
  * This changes the repair swarmbot to handle being able to repair multiple specific items by having eight built-in equipment modifiers of the form "@equipment item 1@" etc.


maw3193:
I force-pushed changes because I noticed a typo in the names of the contragrav swarmbot chassis'

maw3193:
I pushed some more changes because I also noticed that Decontamination Swarms should specialize by Hazardous Materials skill

Library/Ultra Tech/Ultra Tech Equipment Modifiers.eqm
```diff
@@ -45,6 +45,343 @@
...
+						{
+							"id": "fLuL-pD4o6s0lREJs",
+							"name": "Nanobot Dust",
+							"reference": "UT36",
+							"notes": "Immobile",
+							"tags": [
+								"Swarmbots"
+							],
+							"tech_level": "TL11",
+							"cost": "-20%"
```
rinickolous, line 127: This should be -80% instead.

rinickolous (changes requested):
Spotted 1 error but I think this is good to merge as soon as this is addressed.

Library/Ultra Tech/Ultra Tech Equipment Modifiers.eqm
maw3193: aha, thanks for catching that!

I'll fix that now, but I'd like to hold off on merging this, because I've thought of a more flexible way of defining swarmbots that doesn't require an ugly hack to do Multi-Function swarms.
i.e.

* Edit all the existing swarms in Ultra-Tech equipment to note that they're deprecated.
* Create a new equipment item called "@Type@ Swarm".
  - this has an equipment modifier section full of swarm types (e.g. Paramedical Swarm, Repair Swarm). Users can tick multiple if it's a Multi-Function swarm.
  - this has an equipment modifier section full of Microbot chassis' (Microbot Crawler selected by default).
  - this has an equipment modifier section full of Nanobot chassis' (none selected because the default is a microbot crawler)
  - this has an equipment modifier section full of power supplies (TL10 power cells selected by default)
  - this has an equipment modifier for Disguised
  - maybe also add the equipment modifiers for a Thermo-Optic Chameleon Surface and Multispectral Chameleon Surface if I can figure out how they're supposed to work.

Library/Ultra Tech/Ultra Tech Equipment.eqp
```diff
@@ -48131,6 +48131,1076 @@
...
 			}
 		},
+		{
+			"id": "eRd4HdGkY3padXW2c",
+			"source": {
+				"library": "richardwilkes/gcs_master_library",
+				"path": "Ultra Tech/Ultra Tech Equipment.eqp",
+				"id": "eqFME0FUFNk7cR-4s"
+			},
+			"description": "@type@ Swarm",
```
rinickolous, line 48141: Change "@type@" to something else like "@Microbot/Nanobot Type@" to make it easier for users to find this item. I was unable to find it when searching for "Nanobot" for example, and I expect uses to likely have the same issue.

rinickolous (changes requested):
Just one more minor change, then I think this is good to merge.

Library/Ultra Tech/Ultra Tech Equipment.eqp
maw3193: Good point!

This should be fixed now.

PR #355, opened 2025-03-29
Add Snákamaðr.gct

daliin:
Add the snákamaðr (snake-folk) template from Serpents of Legend, p. 46.

### DIFF
--- a/Library/Norðlond/Races/Snákamaðr.gct
+++ b/Library/Norðlond/Races/Snákamaðr.gct
@@ -1,0 +1,1025 @@
+{
+	"version": 5,
+	"id": "Bk0u3yZImc7Bcr2HJ",
+	"traits": [
+		{
+			"id": "T1mieWTqcNDx4tJv4",
+			"name": "Snákamaðr",
+			"reference": "NBS46",
+			"ancestry": "Human",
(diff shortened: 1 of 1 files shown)


Library/Norðlond/Races/Snákamaðr.gct
```diff
@@ -0,0 +1,1025 @@
...
+			"name": "Snákamaðr",
+			"reference": "NBS46",
+			"ancestry": "Human",
+			"container_type": "ancestry",
+			"children": [
+				{
+					"id": "taRmy3DzDabdVdEqm",
+					"source": {
+						"library": "richardwilkes/gcs_master_library",
+						"path": "Basic Set\\Basic Set Traits.adq",
```
dripton, line 16: I don't like the double backslashes in the path when we could use nice clean not-doubled forward slashes.  But digging through all the data files it looks like about 2/3 of them have forward slashes and 1/3 of them have doubled backslashes, so I suspect this is just an artifact of whether you created the data file on macOS/Linux or Windows, and GCS deals with it either way, so I guess it's okay.  No need to change it.

dripton (approved):
This looks good to me.

PR #356, opened 2025-04-06
Add files from Enraged Eggplant

richardwilkes:


dripton (approved):
I pulled these down into my User Library and paged through a bunch, and they look solid.  Far from an exhaustive review, but they're home brew anyway, so there's not really much to check against except Eggplant's blog.  I suggest we go ahead and merge these, then we can correct any problems anyone finds later.

PR #358, opened 2025-04-12
Add Deflect 1 and 2 to DFRPG Equipment Modifiers for armor and shields.

dripton:
These modify the cost of the shield or armor piece, and note the conditional defense bonus, but don't actually affect the dodge/block/parry levels shown for the character because the defense bonus is conditional.


richardwilkes (approved):
Seems reasonable, although I had always thought this was an all-or-nothing type of enchantment. This makes more sense.

PR #359, opened 2025-04-27
Add/update equipment that's used in Loadouts: Starship Crew at TL11

maw3193:
While building a library of TL11 equipment templates based on Loadouts: Starship Crew, I noticed a few equipment items that were missing, some items with incorrect details, and some items that could make use of gcs features that have been introduced since that item was created.

The full list of changes is listed below:

Bio-Tech Equipment: Add biotech facilities
High Tech Equipment Modifiers: Move Firearm Quality accuracy bonus into Features
Basic Set Equipment Modifiers: Move Firearm Quality accuracy bonus into Features
Ultra Tech Equipment:
* Mark Domestic Nanomist as LC4
* Add Domestic Nanocleanser
* Fix Sleep Set's TL12 version being marked TL11
* Update Cybersuits
  - Instead of listing every hit location, changes to the gcs data model
    mean they can list "all" locations.
* Add Laser Chemscanners
* Make Tactical Ultrascanners distinct from regular ultrascanners
  - Make them cost 5x regular ultrascanners
  - Include "Tactical" in their name
* Add 'Tactical' to all Tactical Ladars
* Add optional Chemscanner mode to LADAR
* Add hand grenades
* Electrician Toolkits and workshops have no specialization
  - There are no specializations of the Electrician skill.
* Add a superscience option to portable fusion reactors
* Add the 'Memory Scan' modifier to all Ultrascanners
  - This enables the Ultrascanner to scan a computer's memory
* Update monocrys armour to use conditional DR instead of notes
* Update Assault Boots
  - The bonus DR is on attacks to the underside of the foot, not piercing
    and cutting.
  - Applied the conditional DR bonus as an equipment feature instead of a
    note.
* Update Gripboots
  - Include relevant combat statistics from Assault Boots (which required
    creating variants for TL9 to TL12)
  - Make the bonus kicking damage apply to any attack with usage Kick
    instead of only the Kicking skill (technique?)
  - Add Climbing bonus as a skill modifier
  - Add Ice Climbing bonus as a conditional modifier
* include 'computer monitoring' in the names of the Computer Monitoring Gear
  - Without this, they're just called "Mini Scanner" and "Long Range
    Scanner", which makes it hard to tell what they're for
* Electronics Repair Toolkits and workshops require a specialization
* Add Security Scanners


Library/Ultra Tech/Ultra Tech Equipment.eqp
```diff
@@ -18037,28 +18025,233 @@
 			}
 		},
 		{
-			"id": "elI6n8v2PinZcDSub",
+			"id": "eLSIMa6gfZB_ag_HF",
```
rinickolous, line 18028: The ID of Gripboots (TL9) has been changed, which may break existing synchronization. Is this intentional? I can see how this change could be warranted: It forces those seeking re-synchronisation to re-examine whether they have the right TL variant.

Library/Bio-Tech/Bio-Tech Equipment.eqp
```diff
@@ -1,6 +1,202 @@
...
+			"equipped": true,
+			"calc": {
+				"extended_value": 1000000,
+				"extended_weight": "20,000 lb"
+			}
+		},
+		{
+			"id": "eCplQbxqo6i3neSXw",
+			"description": "Standard Genetics Lab",
+			"reference": "BIO16",
```
rinickolous, line 33: Your Bio-Tech reference abbreviation is "BIO" instead of "BT" as used in the rest of the library. Any reason for this?

Library/Ultra Tech/Ultra Tech Equipment.eqp
```diff
@@ -18395,6 +18588,23 @@
 				"extended_weight": "3.3 lb"
 			}
 		},
+		{
+			"id": "e4rnwDw2XGhGPqBZD",
+			"description": "Hand Grenade",
```
rinickolous, line 18593: While the damage efects of hand grenades are not known, the other weapon stats are and can likely be filled in. Any reason not to do this? Also applies to your other 2 grenade entries.

rinickolous (changes requested):
Just a few minor changes and some queries. Good to merge once addressed.

Library/Bio-Tech/Bio-Tech Equipment.eqp
maw3193: Aha, this was some confusion because Loadouts uses the abbreviation "BIO" for bio-tech. Will fix.

Library/Ultra Tech/Ultra Tech Equipment.eqp
maw3193: At the time I was concerned that there doesn't seem to be *any* combat statistics that can be added to it, as (damage aside as that depends on the warhead), throwing distance is determined by the Throwing Distance table on B355.
Looking at the example on how grenades are done in the Basic Set library, though, throwing distance x3.5 is fine unless the wielder has less than 20 Basic Lift.

In addition, I think I could add the damage effects as a list of ranged weapon usages that default to hidden (plus a note to apply the corresponding warhead equipment modifier to the grenade).

maw3193: I've added ranged attacks for every possible warhead as a separate commit at the end of the series, because it may have been overkill, creates a very big diff, and git isn't helpful at displaying what actually changed, so perhaps it shouldn't be included in the pull request.

Library/Bio-Tech/Bio-Tech Equipment.eqp
maw3193: This should be fixed now.

rinickolous (approved):
I'm a little bit iffy about the note being included in the grenade items by default, but I think this is likely the best solution available given the limitations of the software.

Approved and merged.

PR #360, opened 2025-04-29
Added Discworld

incrsnentrpy:
Content not added is listed in the notes.

### DIFF
--- a/Library/Discworld Roleplaying Game/Discworld Roleplaying Game Equipment.eqp
+++ b/Library/Discworld Roleplaying Game/Discworld Roleplaying Game Equipment.eqp
@@ -29,7 +29,7 @@
 					"defaults": [
 						{
 							"type": "skill",
-							"name": "FLAIL"
+							"name": "Flail"
 						},
 						{
 							"type": "dx",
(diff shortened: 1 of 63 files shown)


dripton:
Taking a look.

Library/Discworld Roleplaying Game/Discworld Roleplaying Game Equipment.eqp
```diff
@@ -0,0 +1,4429 @@
...
+						"base": "1"
+					},
+					"strength": "7",
+					"usage": "Swing",
+					"reach": "1",
+					"parry": "0U",
+					"defaults": [
+						{
+							"type": "skill",
+							"name": "FLAIL"
```
dripton, line 32: You have all the weapon skill names in all caps.  This is inconsistent with other eqp files, which have them in title case, like "Flail".

dripton (changes requested):
It's a huge change so I didn't verify every line but it looks good overall.  The one thing I noticed was that you used ALL CAPS weapon skill names in your eqp files, while other gcs files use Title Case.  I would change this to Title Case  (I'm not sure if having them ALL CAPS actually breaks anything, but we should be consistent anyway.)

dripton (approved):

PR #361, opened 2025-05-05
Fix page references in Banestorm and Supers

maw3193:
While checking page references for GURPS Space, I noticed that some Supers templates were also using S<number> to reference pages in GURPS Supers, instead of SU<number>.

In addition, Geognome in the gcs discord mentioned finding references to Banestorm using B<number> in the Banestorm library.
I was unable to find those erroneous references, apart from the Entertainer's Code of Honor (Theatrical) using the wrong reference.


dripton (approved):
These look good.

PR #362, opened 2025-05-11
Martial Arts Skills: add melee attacks to Hammer Fist

maw3193:
The effect of Hammer Fist is that it's a melee attack that does less damage, but has less risk of injuring the hand.

Melee attack damage for a technique can be represented in gcs, so I added it here.

Note that this may behave strangely if you have both the brawling-based Hammer Fist *and* the Karate-based one, but there's very little reason for a character to have both.

### DIFF
--- a/Library/Martial Arts/Martial Arts Skills.skl
+++ b/Library/Martial Arts/Martial Arts Skills.skl
@@ -1336,6 +1336,26 @@
 				"modifier": -1
 			},
 			"limit": 0,
+			"weapons": [
+				{
+					"id": "wtSAXknaYYGXin3Sg",
+					"damage": {
+						"type": "cr",
+						"st": "thr",
+						"base": "-2"
+					},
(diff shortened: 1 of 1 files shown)


rinickolous (commented):
You could use a replacement `@Brawling/Karate@` for the "Defaults to" field to eliminate the duplicate entry.

Doing the same for the name of the name of the field as well as for the weapon default name discriminator would also eliminate the potential problem you mentioned about the edge case where a character uses the technique for both skills.

image

However, I don't immediately see precedent for this type of naming convention anywhere else in the library and would want @richardwilkes to decide whether setting a precedent with such a change would be warranted.
An alternative to that would be to modify the GCS code to recognise the default skill of a technique as its Specialization for the purpose of skill level resolution. At the moment, specifying a specialization effectively eliminated techniques from being considered. In fact, I just raised an issue for GCS to request the addition of this feature.

PR #367, opened 2025-06-02
Add TL11 Starship Crew Loadouts

maw3193:
This adds all the TL11 loadouts from GURPS Loadouts: Starship Crew.

They're not 100% accurate to the book because:
* They erroneously think that swarmbots are weightless (they weigh 2lbs. I submitted errata).
* They treat the cost and weight of equipment as excluding power cells - I could convert them all to containers and add batteries in, but chose not to because it would be a large change to do for every item in the equipment library that has power cells.
* I'm not certain that they're always applying the cost/weight saving for combination gadgets.
* I don't remove surplus energy cells for combination gadgets.
* They use multiplicative cost modifiers and I use additive ones.

I've done combination gadgets by turning one item into a container and then adding the items its combined with inside it, then applied cost/weight modifiers as appropriate.
It does get slightly confusing that an item could be a combination gadget, have some equipment built-in, or be a container of items that could be removed, and that's all represented by equipment containers.

I haven't done the other tech levels because it's a lot of work and I'm not planning to use them in a campaign yet.

I considered add a Chemscanner and Radscanner to the Science Officer and notes for the Superscience Science Officer that the Ultrascanner replaces them because the TL11 Science Officer lacks scanners without the Superscience lens, but I think that goes a bit further than interpreting the book to the best of my ability.

Potentially, these templates could be combined with the starship crew professional templates, which would be simpler for users, but:
* This is an incomplete set of the equipment options for a professional template (it doesn't include TLs other than 11).
* This would require duplicating the Spacer Basic Kit, Computer Workhorse, Planetary Survival, Better Living Through Chemistry and Armed templates for every professional template.

This also includes:
* Remove the source library data from `@Microbot/Nanobot Type@ Swarm` - it shouldn't have any, and having some causes the checker for changes from the source library to get confused.

### DIFF
--- a/Library/Loadouts/Starship Crew/TL11/Armed.gct
+++ b/Library/Loadouts/Starship Crew/TL11/Armed.gct
@@ -1,0 +1,267 @@
+{
+	"version": 5,
+	"id": "Bf4uJBe7g0UlJVUjN",
+	"equipment": [
+		{
+			"id": "EyA672FEYBxmrBher",
+			"source": {
+				"library": "richardwilkes/gcs_master_library",
+				"path": "High Tech/High Tech Equipment.eqp",
+				"id": "E73yMX2SGu09jndtY"
+			},
+			"description": "Retention Holster",
+			"reference": "HT154",
+			"local_notes": "+2 Retain Weapon.",
+			"tech_level": "7",
+			"tags": [
+				"Firearm Accessories"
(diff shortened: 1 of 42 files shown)


dripton (approved):
I don't have this book so I can't verify every detail, but this all looks sane, so approving.

PR #368, opened 2025-06-07
Ultra Tech Equipment: Fix library sync for swarmbots

maw3193:
Somehow, the swarmbots ended up with source library information saying it came from a nonexistent item in the Ultra Tech Equipment library.

Because of this,  it would always treat swarmbots as being out of sync with the source library.


maw3193:
This has been updated to be based on the current data format.

Also, I noticed some more source library references in there and deleted them too.

PR #371, opened 2025-06-16
Added DR penalties to low tech armor

NoOrangeSkitty:
I added DR penalties to low tech armor e.g. light mails DR on the character sheet will display as 3/1 instead of 3

### DIFF
--- a/Library/Low Tech/Low Tech Armor (by location).eqp
+++ b/Library/Low Tech/Low Tech Armor (by location).eqp
@@ -6,7 +6,7 @@
 			"description": "Head",
 			"modifiers": [
 				{
-					"id": "f6Q72YM9NFSimluT0",
+					"id": "foCwP5Hq5jJLRGS_z",
 					"name": "Cane",
 					"reference": "LT106",
 					"notes": "DR 1; Combustible. If DR is penetrated by burning damage, it can catch fire. See Making Things Burn (p. B433); treat the armor material as resistant; Don time: 9 secs; Holdout -1.",
@@ -32,7 +32,7 @@
 					"disabled": true
 				},
 				{
-					"id": "fI67EXKwsJrwtWbTc",
+					"id": "fOApuptUgqidVtbI2",
 					"name": "Cloth, Padded",
 					"reference": "LT103",
 					"notes": "DR 1*; Don time: 5 secs; Holdout -1.",
@@ -58,7 +58,7 @@
 					"disabled": true
 				},
 				{
-					"id": "fnpzUNdTjTjXkaM3m",
+					"id": "fowWS7hoDz6ZOAuby",
 					"name": "Horn",
 					"reference": "LT106",
 					"notes": "DR 3; Don time: 9 secs; Holdout -3.",
@@ -84,7 +84,7 @@
 					"disabled": true
 				},
 				{
-					"id": "f42eeksR5T7SifsYe",
+					"id": "fIu68wg4g8VYaT_S7",
 					"name": "Layered Cloth, Light",
 					"reference": "LT103",
 					"notes": "DR 2*; Don time: 6 secs; Holdout -1.",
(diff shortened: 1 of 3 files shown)


richardwilkes:
The notes should be updated to not explicitly mention things like "-1 DR vs impaling", since that's what this change does for you.

rinickolous:
@NoOrangeSkitty, could you action the requested change? I'm happy to merge after that's done.

Library/Low Tech/Low Tech Armor (by location).eqp
```diff
@@ -165,7 +165,7 @@
 					"id": "f0GzvAvrcbGT6AtMZ",
 					"name": "Leather, Medium",
 					"reference": "LT104",
-					"local_notes": "DR 2*; -1 DR vs. impaling; Don time: 9 secs; Holdout -1.",
```
rinickolous: Good job on removing the "-1 DR vs. X" notes, but this equipment now does not reflect the damage penalty mechanically. The "DR 2" and other such notes should also be removed AFAICT, given those features *are* implemented mechanically.

Library/Low Tech/Low Tech Armor (by material).eqp
```diff
@@ -2013,12 +2027,21 @@
 					"disabled": true
 				},
 				{
-					"id": "fvK3nbMPU3Q1zHb7v",
+					"id": "f1KIqw5RoCX_J9scJ",
```
rinickolous: These ID changes are also reflective of these objects being in a library other than the master library. I don't think it matters for Equipment Modifiers, fortunately.

Library/Low Tech/Low Tech Instant Armor.eqp
```diff
@@ -5426,14 +5570,16 @@
 					"locations": [
 						"abdomen"
```
rinickolous: inconsistency for IDs. "abdomen" vs "groin"

Library/Low Tech/Low Tech Instant Armor.eqp
```diff
@@ -6573,14 +6870,18 @@
...
 					],
-					"amount": 4
+					"specialization": "cr",
+					"amount": -2
 				},
 				{
 					"type": "dr_bonus",
 					"locations": [
-						"face"
+						"face",
```
rinickolous: "neck" is missing

Library/Low Tech/Low Tech Armor (by location).eqp
rinickolous: These features are outdated, using 1 feature per location. I think you've just removed the comment here and have not updated the features to reflect the DR this provides.

NoOrangeSkitty:
The mistakes in item descriptions were already there, as i used notepad++ to replace all of the "- dr vs x" parts of the description all at once. I also forgot to update my Gcs version so that probably caused issues

rinickolous:
> The mistakes in item descriptions were already there

That's fine, you don't need to address that since the scope of this PR doesn't necessitate it. However, there are several other issues which need to be addressed, as I pointed out above.

Library/Low Tech/Low Tech Armor (by location).eqp
```diff
@@ -7248,337 +6752,337 @@
 			"local_notes": "Roll 1d; on 6, the armor is hit.",
```
rinickolous: Not sure what we should do with armor that doesn't have a feature attached to it at all (because the hit location is not included by default in GCS). It may be justifiable to include DR in the notes in such instances, but I'd want to get a second opinion. @richardwilkes , @dripton , thoughts?

NoOrangeSkitty:
Should i add dr in the description of partial armor

Library/Low Tech/Low Tech Instant Armor.eqp
rinickolous: neck is still missing

Library/Low Tech/Low Tech Instant Armor.eqp
```diff
@@ -10230,13 +10769,17 @@
 				{
 					"type": "dr_bonus",
 					"locations": [
-						"torso"
+						"groin",
```
rinickolous: "groin" is present in this feature but not the other one.

NoOrangeSkitty:
i havent actually finished instant armor yet, that is the old version

PR #373, opened 2025-07-06
Add Bio-Tech Meta-Traits

maw3193:
While looking into adding the Bio-Tech racial templates, I found some more traits from Bio-Tech that I hadn't added.

In this Pull Request, I went through Bio-Tech's Gengineered Traits and added the meta-traits and features I found that didn't seem to have an equivalent in Basic Set.

This includes building a few abilities based off innate attacks and afflictions, where I've tried to use the existing traits and modifiers to make most of them syncable with the source library.

Most of the changes are:
* Change modifiers to take substitutions and set its % value by setting its level.
* Add missing modifiers.

### DIFF
--- a/Library/Basic Set/Basic Set Enhancement Modifiers.adm
+++ b/Library/Basic Set/Basic Set Enhancement Modifiers.adm
@@ -273,9 +273,16 @@
 		},
 		{
 			"id": "mcXN1MCeZnZB4Jz2e",
-			"name": "Follow-Up",
+			"name": "Follow-Up (@Carrier Attack@)",
 			"reference": "B105,P102",
-			"local_notes": "Cost is variable and must be entered manually"
+			"local_notes": "1 point per level",
+			"cost": 1,
+			"levels": 1
+		},
+		{
+			"id": "m1nJdcOiKLxg7pXwe",
+			"name": "Follow-Up (@Carrier Attack@)",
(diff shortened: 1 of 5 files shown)


maw3193:
I'm slightly concerned about one of my changes - I changed Follow-Up from being worth +0% and having a note to manually edit the point value, to being worth 1%, intended to be used as 1% per level,  but if you set the level to 0 it's still worth 1%, and 9.01 points will get rounded to 10 points.

maw3193:
> I'm slightly concerned about one of my changes - I changed Follow-Up from being worth +0% and having a note to manually edit the point value, to being worth 1%, intended to be used as 1% per level, but if you set the level to 0 it's still worth 1%, and 9.01 points will get rounded to 10 points.

I resolved this by adding a variant of Follow-Up worth +0% that doesn't have levels.
None of the other modifiers I made into levelled ones would be used as +0% as far as I can tell.

Also, this has been updated to use the new data format.

maw3193:
The most recent change adds the Taboo Traits features from bio-tech to the trait library.

maw3193:
The most recent change adds the No Appendix feature to Bio-Tech traits, adds Increased Size Modifier and Decreased Size modifier to Basic Set Traits, and alters the Payload trait to calculate the payload's capacity based on the trait's level and the character sheet's Basic Lift.

PR #374, opened 2025-07-15
Characters from Discworld GURPS game

Verny56:
Here are some characters from the book to add to the existing library for Discworld.


richardwilkes:
I'm unclear why the traits were broken into three separate files, unlike everywhere else in the Master Library.

Library/Discworld Roleplaying Game/Characters/Adora Belle Dearheart.gcs
```diff
@@ -0,0 +1,2600 @@
...
+		"age": "25",
+		"birthday": "January 29",
+		"eyes": "Brown",
+		"hair": "Black",
+		"skin": "Pale",
+		"handedness": "Right",
+		"gender": "Feale",
+		"height": "5'7\"",
+		"weight": "130 lb",
+		"player_name": "LAPTOP-THEGOA5C\\Wayne",
```
rinickolous: This player name should be removed from all character files.

rinickolous (changes requested):
these files use an old format which does not include some modern GCS features. There are also some other issues which I pointed out in comments from an initial look. As Richard mentioned, the Traits should also not be split into 3 separate files, but should be one file. Please update this branch, we can't merge it in its current state.

rinickolous:
Closing this as inactive.

PR #379, opened 2025-07-27
Bio-Tech: Add Gengineered Human Racial Templates

maw3193:
This adds all the Gengineered Human Racial Templates from Bio-Tech.

They are implemented as Templates that use Template Choices to choose either the base template or one of the Sub-Races (where they exist).

This is based on the work by anders3865 on the gcs discord, change to support the aforementioned Template Choices, plus the traits are linked to ones in source libraries, and the template's TL, cost and LC are listed.

This uses traits from #373 so if this pull request is merged before that one then gcs will report that the traits differ from their source libraries (or that the source library is missing).


dripton (approved):
I didn't see any issues

PR #384, opened 2025-07-30
Add Gengineered animals from Bio-Tech

maw3193:
This contains:
*  All the sapient sample gengineered animals from Bio-Tech.
* It adds source library references to the Quadruped and Domestic Animal meta-traits
* Adds Increased/Decreased Air/Water Move traits. This has no features, because I'm not confident about adding Ground, Air and Water move attributes to all character sheets.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -5438,6 +5438,23 @@
 			}
 		},
 		{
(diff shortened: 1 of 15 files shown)


maw3193:
Some of these templates also have equivalents from Transhuman Space, but the one I did check has different traits.

dripton (approved):
These look good.

PR #386, opened 2025-08-01
Sorted the DR trait's modifiers to match book order + added a reference note to Hardened

cognato-net:
Sorted the modifiers to match the order in the books:
1. Enhancements first, then Limitations;
2. Inside each category, alphabetical order;
3. For different sub-types of the same modifier, I matched the book order.

For simplicity, rather than adding separate categories for each of the other books (Horror, Psi, etc.), I placed them inside those same two major categories (Enhancements and Limitations), in alphabetical order.

For Hardened, I also added a note with a reference to the armor divisor steps, which is VERY handy if you're playing a super with Hardened DR.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -5048,7 +5048,7 @@
 		{
 			"id": "tBmSJ3pNE53c4IRUU",
 			"name": "Damage Resistance",
-			"reference": "B47,P45,MA43,PSI14",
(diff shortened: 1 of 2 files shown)


cognato-net:
This is a separate change request from the previous Damage Resistance one. Tooltips don't render HTML, so I removed HTML tags from tooltips in various .attr files inside Settings/Attributes.
I understand that GitHub exists precisely to prevent version conflicts, so I trust that everything works as intended, but I'm not very familiar with how this works so this is just to clarify that they're two separate things.

PR #388, opened 2025-08-01
Sorted Damage Reduction's modifiers per book order + added a reference note to Hardened

cognato-net:
Sorted the modifiers to match the order in the books:
1. Enhancements first, then Limitations;
2. Inside each category, alphabetical order;
3. For different sub-types of the same modifier, I matched the book order.

For the non-Basic Set modifiers, rather than adding separate categories for each of the other books (Horror, Psi, etc.), I placed them inside those same two major categories (Enhancements and Limitations), in alphabetical order.

And, for Hardened, I added a note with a reference to the armor divisor steps, which is VERY handy if you're playing a super with Hardened DR.


dripton (approved):
This is alphabetizing and adding references, all good.

PR #390, opened 2025-08-03
Added "Based Cosmo" folder inside the Homebrew folder

ggnoobed:
I wanted somewhere to put some of my stuff for easier access from my players, as well as an enchantment library from GURPS Magic and GURPS DF8.


dripton (approved):
It's homebrew, so didn't check very closely, just made sure things looked reasonable.

PR #396, opened 2025-08-08
Added all 3 classes, traits, and skills from DF20

cognato-net:
Demon-Slayers, Mage-Slayers, and Undead-Slayers, including their respective Traits, Skills, Perks, and Power-Ups.

This fulfills #219.


cognato-net:
> For the prerequisites, unless there is a need to allow for specialties, name matching should probably use "whose name is", not "whose name contains". This makes it less likely to match things by accident and is also slightly better performing when resolving these things.

There we go, should be fixed now.

Note: While making these adjustments, I found that the Basic Set skill "Research" (which is included in the Demon- and Undead-Slayer templates) has some curiously formatted requirements:

image

* Is the lack of closing parenthesis intentional? I don't see why, given that it already matches on "contains".
* Shouldn't all three of these be rolled up into just one single "contains" "written" requirement?

richardwilkes:
I'm busy with some other stuff right now, so not sure when I'll get a chance to take a look at this again. If another week goes by without one of us either approving or commenting, please poke me again.

dripton (approved):
Looks reasonable to me

PR #397, opened 2025-08-08
Adding Starting Wealth per B27 and B25

cognato-net:
Can we add this to the Basic Set Equipment library? I think it would be very helpful for many players.
PS: This is all taken verbatim from the "Tech Level and Starting Wealth" table on B27 and the "Wealth" trait on B25 -- there's no homebrew here.


cognato-net:
Should be cleared now.

richardwilkes:
Trying this out and was thinking this would likely be better as just a Note whose content automatically calculated the remaining starting wealth you should have available to you. With some minor scripting, it wouldn't require any input from the user -- just drop it into the sheet and it could look for the relevant traits that modify wealth level, lookup the TL from the sheet, compute the total spent already, and give you a remaining value... assuming that's what this was intended for.

rinickolous (approved):
I see no issue with this. I wouldn't really use this myself, but I suppose it makes sense.

dripton:
I think starting wealth is something the GM picks, like point level.  So I don't think it makes sense to have GCS compute it unless there's an easy way to override that number.  That said, it's not that hard to ignore either, so meh.

PR #399, opened 2025-08-09
Stress & Derangement From GURPS Horror

trechriron:
1. Created a new attr file.
2. Tested locally on GCS.

### Overview
This Attribute file contains two reference-only pools: Stress and Derangement. They each include one threshold as per the rules in Horror on pg. 142. Instead of tracking these points as the book suggests, it is easier to track them as positive points to 0 (vs. 0 points into negatives).

Tooltips were included.

Stress includes "Calm" down to 0, then "Overstressed".

Derangement includes "Sane" down to 0, then "Disturbed".

It is possible to enter negative numbers. Which is not terrible since they are there for tracking only.

### DIFF
--- a/Library/Settings/Attributes/Stress and Derangement.attr
+++ b/Library/Settings/Attributes/Stress and Derangement.attr
@@ -234,7 +234,7 @@
 			"base": "$will",
 			"thresholds": [
 				{
-					"state": "Over Stressed",
+					"state": "Overstressed",
 					"value": "0",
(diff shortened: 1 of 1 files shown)


richardwilkes:
Now that v5.38.0 introduced the ability to merge/append attribute data, it would probably be better to remove the standard stuff from this and just leave the two new attributes.

trechriron:
Updated.

PR #406, opened 2025-08-24
Fix Quarterstaff in Basic Set Equipment

GriffinLore:
The Quarterstaff's Melee Usage stats included stats for swords which is not aligned to the stats published in B273. I removed the two additional usage stats.


GriffinLore:
First commit had a few extra unintended changes.

GriffinLore:
You’re absolutely right... thanks for pointing that out. I wasn’t thinking of the quarterstaff in the context of sword usage and had overlooked B274 where it specifies the Two-Handed Sword skill. I’ll revert my change so the additional usage stats remain accurate. Appreciate the correction!

PR #407, opened 2025-08-25
Fixes to Psionic Powers

maw3193:
The changes to psionic powers are:
* Fix PK Shield missing its DR feature
* Turn TK Grab and Telekinetic Control into trait containers, containing the Telekinesis advantage
  - This is so that telekinetic thr and sw damage can detect it.
* Make Astral Sword use the levelled and non-levelled damage instead of the workaround.
* Fixed the Astral Perception psionic package accidentally using the Astral Sword skill instead of Astral Sight.
* Update the templates that use these skills and abilities.

I haven't touched the Enraged Eggplant psionic powers, as they seem to already be built in a way that is compatible with telekinetic sw and thr.

### DIFF
--- a/Library/Action/Action Traits.adq
+++ b/Library/Action/Action Traits.adq
@@ -839,6 +839,7 @@
 					"weapons": [
 						{
 							"id": "wkgRi94gFhs__aoLu",
+							"sv": 1,
 							"damage": {
 								"type": "cr",
 								"st": "thr",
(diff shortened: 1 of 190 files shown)


rinickolous (approved):
GCS seems to be inserting a double space into the `.calc.damage` property where an armor divisor is not present, but this isn't an issue with the PR and doesn't affect the data in a way that matters within GCS, so no issue there.

Approved.

PR #412, opened 2025-08-30
Fix Enhanced Parry (@One weapon skill@) to add a weapon parry modifier

maw3193:
This affects the Enhanced Parry traits in:
* Basic Set Traits
* Action Traits
* DFRPG Traits
* Discworld Roleplaying Game Traits
* Lite Traits

and updates all the templates and characters that use Enhanced Parry to use the fixed version.


dripton (approved):
There's a lot here, but it looks consistent.

PR #413, opened 2025-09-02
Pyramid 69: Add the Mind and Body article

maw3193:
This includes everything in the Pyramid 3-69: Mind and Body article,
except for the rules in [New Psi-Amplifier](PY69:15) for creating
psi-amplifiers with a higher or lower bonus. Instead the finished
psi-amplifier amulet, armband, gauntlet and hilt have been added to The
Way of the Cerulean Blade's template equipment.
    
With thanks to Enraged Eggplant, whose homebrew psionics library saved
me from creating some stuff from scratch.

This MR is built on top of #411 and should be merged after it.


maw3193:
I think my changes to the Telekinesis and Psychokinesis Talent aren't ready to be merged. The talents also cover imbuement skills that shouldn't be tagged with a psionic power by default.

maw3193:
The latest version of this pull request should no longer delete imbuements' skill bonus.

maw3193:
The latest push replaced the change to Psychokinesis Talent to just add the Psi-Sword and Psi-Shield skill bonus.

I also changed the tags on some of the new skills - I think Psi-Sword and Psi-Shield shouldn't be tagged as Telekinesis (in the same way that PK Shield isn't Telekinesis).

rinickolous (changes requested):
As per your request, #415 is now merged. Please make the necessary changes you mentioned in that PR to this branch and re-request a review when ready.

PR #414, opened 2025-09-05
Fix tags on psionic talents

maw3193:
Broadly, this changes psionic talents to use tags instead of listing every psionic skill by name, and changes the tags of some psionic skills to make sure this is handled correctly.

I believe that these talents were added at a time when gcs didn't support adding a skill bonus to any skill with a particular tag.

Adding the skill bonus by tag is also convenient for pyramid/third-party/homebrew/player-created psionic abilities being added to existing powers.

### DIFF
--- a/Library/Psionics/Psionics Skills.skl
+++ b/Library/Psionics/Psionics Skills.skl
@@ -2945,8 +2945,7 @@
 					"reference": "PSI41",
 					"tags": [
 						"ESP",
-						"Psionics",
-						"Remote Senses"
+						"Psionics"
 					],
 					"difficulty": "per/h",
 					"defaults": [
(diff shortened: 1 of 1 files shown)


Library/Psionics/Psionics Skills.skl
```diff
@@ -2945,8 +2945,7 @@
 					"reference": "PSI41",
 					"tags": [
 						"ESP",
-						"Psionics",
```
Calmquist, line 2948: Why are you removing the Psionics and Remote Senses tags?

maw3193:
> How are you making sure techniques are not double counted? I tried using tags before, but I ran into trouble with techniques.

Hm yes, you're right. I think I must have misunderstood what was going on when I first tested it.

Withdrawing this Pull Request and rethinking my options.

PR #420, opened 2025-09-13
Add extra defaults for Armoury specialties

Enzarel:
Added TL-dependent defaults to Smith (Bronze), Smith (Iron), and/or Machinist for Armoury (Body Armour), (Melee (Weapons), (Small Arms), per B178

Also fixed the default for the generic (@ Armour type@) version to actually use the chosen specialty when defaulting to Engineer.

### DIFF
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -814,7 +814,7 @@
 				{
 					"type": "skill",
 					"name": "Engineer",
-					"specialization": "Battlesuits",
+					"specialization": "@Armor type@",
 					"modifier": -4
 				}
 			],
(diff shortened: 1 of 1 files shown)


Library/Basic Set/Basic Set Skills.skl
```diff
@@ -868,6 +868,35 @@
...
+						"qualifier": 1
+					}
+				},
+				{
+					"type": "skill",
+					"name": "Smith",
+					"specialization": "Iron",
+					"modifier": -3,
+					"when_tl": {
+						"compare": "at_most",
```
dripton, line 888: Technically this should be TL2-4 rather than TL <= 4 but I don't think GCS supports that currently, so this is probably the best we can do.

dripton (approved):
I think this looks good.  It would be nice to be able to say TL2-4 rather than TL<=4 for the iron age stuff, but since we currently can't do that, this is the best we can do.

PR #421, opened 2025-09-13
Add HT58-59 Survival Kits

Enzarel:


Library/High Tech/High Tech Equipment.eqp
```diff
@@ -37720,6 +37741,153 @@
...
+				"extended_value": 6,
+				"weight": "1 lb",
+				"extended_weight": "1 lb"
+			}
+		},
+		{
+			"id": "eJ5Vp6njReziLM4X9",
+			"description": "Personal Basics",
+			"reference": "HT58",
+			"local_notes": "Status 1; @Environment@ Survival specialty.",
```
dripton, line 37790: Interesting that you split them up by Status since Status affects Cost of Living.  I probably would have just used the Status 0 cost of living to price Personal Basics.  I think it's fine though; if a GM wants to require that high-status people buy more expensive Personal Basics they can, and if they don't want to they can just let everyone use the Status 0 version.

dripton (approved):

PR #422, opened 2025-09-14
Add prereq count to Magic Spells

dripton:
I just used the numbers from the tables on M223 to M237.  I didn't hand calculate anything.  I left out the ones that said "Varies" since it needs to be a number.

### DIFF
--- a/Library/Magic/Magic Spells.spl
+++ b/Library/Magic/Magic Spells.spl
@@ -19,6 +19,8 @@
 			"maintenance_cost": "Varies",
 			"casting_time": "2 sec",
 			"duration": "1 min",
+			"prereq_count": 20,
+			"prereq_count": 2,
 			"prereqs": {
 				"type": "prereq_list",
 				"all": true,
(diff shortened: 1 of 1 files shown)


Library/Magic/Magic Spells.spl
```diff
@@ -19,6 +19,8 @@
 			"maintenance_cost": "Varies",
 			"casting_time": "2 sec",
 			"duration": "1 min",
+			"prereq_count": 20,
+			"prereq_count": 2,
```
richardwilkes, line 23: This looks broken.

dripton: I did a few inside GCS, but it was too slow, so I did the rest in vim where I had macros, copying numbers from the tables in the back of Magic.  I loaded the file into GCS when I finished to make sure it was still well-formed JSON but clearly missed that double-entry above, sorry.  (I wouldn't be surprised if there were a couple more typos as this was hundreds of numbers.)

dripton: The new "id" and "sv" stuff under weapons were added by GCS not me.

richardwilkes (approved):

Library/Magic/Magic Spells.spl
richardwilkes: Yeah, the 'sv' field is new-ish and expected. The changing IDs are not, but I haven't yet tracked down why that happens sometimes.

PR #452, opened 2026-01-20
Bio-Tech: Add Biogadgets

maw3193:
This includes the Biogadget modifiers, the nutrient solution that's implied to be purchasable, and the example biogadgets.


Library/Bio-Tech/Bio-Tech Equipment.eqp
```diff
@@ -215,6 +215,314 @@
...
+			"calc": {
+				"value": 2000,
+				"extended_value": 2000,
+				"weight": "0 lb",
+				"extended_weight": "0 lb"
+			}
+		},
+		{
+			"id": "et-4O1teFAHna0qBZ",
+			"description": "Biofilter Canteen",
```
rinickolous: Shouldn't the Biofilter Canteen be a container by default?

PR #453, opened 2026-01-20
fix: HTML - NPC card add border bottom padding to show dr

opcode0x1B6:
Adds a 0.3 inch padding to show the DR/armor table previously cut off by the bottom border in the html npc card export.
Tested in Firefox.

New bottom border of the NPC card

New bottom border of the NPC card,

Old bottom border of the NPC card

Old bottom border of the NPC card.


opcode0x1B6:
Could not get the box-sizing to work to correct the aspect ratio.
I'll just use chromium to render the cards instead of trying to fix it for firefox.
Thanks for your feedback.

PR #468, opened 2026-01-26
Fix issue with Silence trait feature bonuses

LeeSaferite:
The Silence trait features provide bonuses to the Stealth skill that scale with level. This fixes most instances where Silence was applying a Stealth bonus as a fixed amount.

NOTE: The Silence trait still needs another pass across all locations to bring it into sync with how the Basic Set represents the trait now. The direct Stealth bonus is not 100% correct and is invisible if the Stealth skill is not on the character sheet. It also does not convey the difference in bonus amount between moving vs not moving. The latest Basic Set version of the trait has a note explaining that it's only applicable for hearing based Stealth checks and it communicates the bonuses by using conditional modifiers that show even without the Stealth skill being on the character sheet.

### DIFF
--- a/Library/Banestorm/Races/Halfling.gct
+++ b/Library/Banestorm/Races/Halfling.gct
@@ -140,7 +140,8 @@
 								"compare": "is",
 								"qualifier": "stealth"
 							},
-							"amount": 2
+							"amount": 2,
+							"per_level": true
 						}
 					],
 					"can_level": true,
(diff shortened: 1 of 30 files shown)


rinickolous (commented):
Going to delay this slightly and make sure the library is all on the same page regarding how Silence should apply its bonus. Bear with.

rinickolous:
I don't think a skill bonus is the way to handle this, should be a Conditional Modifier instead as you mentioned in your original comment. If you could modify this PR to change it to that, that would be great.

rinickolous:
Closing this as inactive. As I stated earlier, I think this would be better handled with a Conditional Modifier. Please open a new PR if you would like this reviewed again.

PR #476, opened 2026-02-01
Power Ups: Make the Incompetence quirks easier to use

maw3193:
Players may be incompetent at an important skill, and so would benefit from seeing what their skill level is at default, which gcs can show if you have that skill at 0 points.

Previously, having the skill on your sheet at all would fail Incompetence's prerequisites, so I replaced it with a script to check whether the skill has any points in it, instead.

This also includes showing the skill penalty from Incompetence, as it can be shown on the character sheet now.

For completeness, I also added the 0 skill points invested prerequisite to Delusional Competence.

### DIFF
--- a/Library/1shotadventures/Characters/Who Tracks the Steps of Glory/Michael_Westinghouse.gcs
+++ b/Library/1shotadventures/Characters/Who Tracks the Steps of Glory/Michael_Westinghouse.gcs
@@ -767,14 +767,32 @@
 			}
 		},
 		{
-			"id": "twTYD_R47pXUgGjbN",
-			"name": "Incompetence (Piloting)",
+			"id": "tUFwuOtI2D7DGeVY9",
+			"source": {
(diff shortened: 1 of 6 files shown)


maw3193:
> A prereq that reads: "doesn't have a skill whose name is @Skill@ and whose specialization is anything and whose level is at least 0" should already do the right thing -- and does in my testing just now with the current version of GCS. It wouldn't surprise me if the prereq wasn't setup right for that, but there should be no need to use a script, which should be avoided if at all possible, as the overhead for them is significant.

I've tried that, and it seems to only work if the skill's level without any points in it is 0 or less.

Here's the skill with its prerequisites set as you say:
image

Here's a character sheet with the skill greater than 0:
image

Here's a character sheet with the skill at 0:
image

it looks like that prerequisite checks the skill's level is greater than 0, not whether it has more than 0 points invested in it.

PR #481, opened 2026-02-15
Update Low-Tech Instant Armor and list them by type and by material

maw3193:
This includes the following changes submitted by Teragen on Discord:
* Equipment is categorised using equipment containers. They are split by either type or by coverage, and then type and coverage subdivide by the various types or hit locations.
* Split DR is encoded into the "DR bonus" feature, instead of being purely based on notes.
* Some missing equipment from Instant Armor (like the Arming Doublet) have been added too.

I have made the following changes to that submission:
* Deleted stray source library references that came from the submitted file being created inside the user library.
* For any items that had their IDs changed, I changed them back.
* Because listing items in two different ways means duplicating them, I added source library references to the duplicate items referring to the originals. GCS currently doesn't seem to sync the items together if they both exist in the same library - which unfortunately means that altering the source item doesn't let it be automatically synced to the duplicate one - but since the source library information points to the original, once it's added to any other file, it can correctly pull changes from the original.

### DIFF
--- a/Library/Low Tech/Low Tech Instant Armor.eqp
+++ b/Library/Low Tech/Low Tech Instant Armor.eqp
@@ -21,7 +21,8 @@
 					"locations": [
 						"abdomen",
 						"arm",
-						"torso"
+						"torso",
+						"vitals"
 					],
 					"amount": 0
 				}
@@ -52,7 +53,7 @@
 					"type": "dr_bonus",
 					"locations": [
 						"groin",
-						"hand"
+						"leg"
 					],
 					"amount": 0
 				}
--- a/Library/Low Tech/Low Tech Instant Armor.eqp
+++ b/Library/Low Tech/Low Tech Instant Armor.eqp
@@ -5375,7 +5375,7 @@
 			"description": "Leather, Heavy Groin Armor",
 			"reference": "LTIA7",
 			"reference_highlight": "Leather, Heavy",
-			"local_notes": "Don: 3 secs; Holdout: -4; Reaction Pen.-1; Flexible and susceptible to blunt trauma (p. B379); -1 DR vs. impaling.",
+			"local_notes": "Don: 3 secs; Holdout: -4; Reaction Pen.-1; Flexible and susceptible to blunt trauma (p. B379)",
 			"tech_level": "0",
 			"tags": [
 				"Groin Only",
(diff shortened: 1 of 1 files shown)


rinickolous:
@maw3193 lmk when ready

Library/Low Tech/Low Tech Instant Armor.eqp
rinickolous, line 1017: I don't think Cane Bascinets exist in LTIA.

Library/Low Tech/Low Tech Instant Armor.eqp
```diff
@@ -1264,25 +1346,30 @@
 			"id": "e2htbNlX1xn4-u6x7",
 			"description": "Cloth, Padded Bascinet, Padded",
```
rinickolous, line 1346: double padded?

Library/Low Tech/Low Tech Instant Armor.eqp
```diff
@@ -1302,7 +1389,11 @@
 			"local_notes": "Don: 11secs; Holdout: 0; Flexible and susceptible to blunt trauma (p. B379)",
```
rinickolous: missing space

rinickolous (changes requested):
there are lots of foot items (boots, sollerets, etc.) that have a trailing space in the name.
several items have LC 5, which is incorrect. some items seem to be the result of duplication which could be more careful, or of algorithmic duplication of some kind as far as I can tell.

Let's start with removing non-existent items, fixing the pointed out typos, and just double checking the names and notes of everything once.

Library/Low Tech/Low Tech Instant Armor.eqp
maw3193: I think they do. my copy of Instant Armor has Bascinets on page 14, and Cane is right at the top of the table. I'm not aware of any errata for Instant Armor that might remove it.

Library/Low Tech/Low Tech Instant Armor.eqp
maw3193: Apparently so. It's made of Padded Cloth, and because it's a helmet you can also add padding.

Library/Low Tech/Low Tech Instant Armor.eqp
maw3193: Agreed. will fix.

maw3193:
For LC, I'm finding contradictory guidance in Basic Set and Low-Tech.
Basic Set gives Low-Tech Armor no LC if it's a clothing item that could provide some DR (e.g. boots, gloves, leather trousers); LC4 for bronze, leather or light mail armour; and LC3 for anything more serious than that.

Low-Tech says to give is LC4 or LC3 if it could plausibly be categorised as clothing but provides some protective benefit, and LC2 or LC1 if it would only be worn for combat.

For the moment, I'll make sure that clothing items have no LC and that LC5 is replaced with a more sensible number.

maw3193:
I've pushed changes to tidy up the names and descriptions, and the misapplied LC5.

I haven't noticed any equipment items that shouldn't be there.

maw3193:
The latest push was just rebasing against the latest commit in the main branch

PR #482, opened 2026-02-17
Add documentation for `$WORD` argument for external PDF viewer

justanothercell:
This adds documentation for richardwilkes/gcs#1024

### DIFF
--- a/Library/Markdown/User Guide/General Settings.md
+++ b/Library/Markdown/User Guide/General Settings.md
@@ -64,7 +64,7 @@ GCS will automatically replace:
 
 - `$PAGE` with the referenced page number.
 - `$FILE` with the correct PDF file, based on the **Page Reference Mappings** in [Page References](Page%20References).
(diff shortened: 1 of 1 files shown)


Library/Markdown/User Guide/General Settings.md
```diff
@@ -64,6 +64,11 @@ GCS will automatically replace:
 
 - `$PAGE` with the referenced page number.
 - `$FILE` with the correct PDF file, based on the **Page Reference Mappings** in [Page References](Page%20References).
+- `$WORD` with the highlighted word.
```
richardwilkes: Should probably be "highlighted text", since it can be (and often is) more than a single word.

richardwilkes: Hmm... I wonder if the variable name should also be changed for the same reason? Perhaps to `$TEXT` or `$HIGHLIGHT`?

PR #483, opened 2026-02-19
Pyramid 103: Add the Mad as Bones article

maw3193:
As requested by Kungliao on the GCS Discord.

This is an article about implementing Horror's Stress and Derangement as a pool like Fatigue Points and Long-Term Fatigue Points.

This includes:
* The attributes added by the article, Stability Points and Long-Term Stability Points, as a new attributes file.
* The new traits added by the article
  * Traits to increase/decrease the new attributes
  * Grounded and Unhinged
  * A new Stability Attack version of Innate Attack
* The new modifiers added by the article
  * New trait modifiers added to Damage Resistance, Healing, and Terror.
  * New trait modifier Destabilizing, that applies to any disadvantage that takes a self-control rooll.
  * New trait modifiers Drains Stability Points and Inflicts Long-Term Stability Points.

This does not include:
* The example abilities from the article, as they are quite complicated.
* A Specific Resistant trait for Resistant to Mental Stress, as that is already doable with the existing trait.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,4 @@
 * @richardwilkes
 
 # Library custodians
-/Library @richardwilkes @rinickolous @ethorad @dripton
+/Library @richardwilkes @rinickolous @dripton @maw3193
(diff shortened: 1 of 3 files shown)


rinickolous (approved):
Looks good.

PR #485, opened 2026-02-23
Pyramid 65: Automated Natural Weapons customization, mostly, and added Sample Abilities.

Enzarel:
Previously, Natural Weapons from PY65:24-26 had no pre-made weapon profiles or features; all stats and modifiers had to be applied manually by the user.

Here, all chosen modifiers are applied automatically to the pre-made attached weapon, with the following exceptions in two categories:

**Type 1**, which are typically described only in the trait/item description, and _not_ described on the weapon profile, or reflected in stats:
- Destructive Parry[^1]
- Disarming[^2]
- Hidden[^3]
- Single[^4]

**Type 2**, which are impossible to implement at present:
- Extra damage type[^5]
- Projected[^6]
- Ranged[^7]
- Swing-capable (All types)[^8]

As a workaround: extra weapon profiles (hidden by default, that have to be manually un-hidden by the user) are provided to facilitate Ranged and Swing-Capable. To facilitate Projected, the Ranged attacks default to `Brawling` with Specialization `<delete this if you took Projected>`+0

To allow easy choosing of the primary damage types, the old singular Natural Weapons trait was split into 11. Cutting variant (with sample modifiers, all automatic) seen below.

image

I also added the remaining Sample Abilities from the article:
- Dragon's Breath, Freezing Strike, Hammer Hands, Lightning Whip, Minotaur's Horns, Psychic Dagger, Telekinetic Push, Ultra-Sharp Claws

...and added weapon profiles to the existing ones:
- Bone Stakes, Chi Blast, Decaying Touch

[^1]: Always damages opponent's weapon on contact.
[^2]: +2 to disarming attempts.
[^3]: Can be drawn or retracted as a free action.
[^4]: Comes in one copy, instead of one per limb.
[^5]: Adds a secondary usage with any chosen damage type.
Sadly, weapons can't be added, unhidden, nor their damage types changed by Features.
[^6]: Allows Ranged uses at Brawling instead of Innate Attack.
Sadly, skill defaults can't cleanly be changed by Features.
[^7]: Adds a Ranged usage.
Sadly, a weapon can't be added nor unhidden by Features.
[^8]: Changed damage from `thr` to `sw`.
Sadly, neither a weapon's hidden status nor damage source can be changed by Features.

### DIFF
--- a/Library/Pyramid Magazine/Pyramid 65/Pyramid 65 Traits.adq
+++ b/Library/Pyramid Magazine/Pyramid 65/Pyramid 65 Traits.adq
@@ -6768,21 +6768,21 @@
 					],
 					"modifiers": [
 						{
-							"id": "mGQr_ssihpiE1IgZ9",
+							"id": "mG9ORvJgauGwk2Ze4",
 							"name": "Destructive Parry",
 							"reference": "PY65:24",
 							"cost_adj": "40%",
 							"disabled": true
 						},
 						{
-							"id": "md7yBFhroa-sefT_R",
+							"id": "m8PD4VCJ1qWq6ZF6V",
 							"name": "Disarming",
 							"reference": "PY65:24",
 							"cost_adj": "20%",
 							"disabled": true
 						},
 						{
-							"id": "mp3ybOqiNvjTA_gxL",
+							"id": "m4IeSAsV30toCsCBu",
 							"name": "Extra Damage Type",
 							"reference": "PY65:24",
 							"local_notes": "***Apply manually!***",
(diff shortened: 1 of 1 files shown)


maw3193:
Also, I spotted your comment apologising for the trait modifier IDs changing every time you edit it.

That's something I've also experienced - when I create commits to push, I use `git add -i` to interactively pick every change that goes into the commit, so that I can ignore the ones that just change IDs, and then I use `git checkout HEAD -- <file path>` to replace the contents of the file with the committed version.
Though I wouldn't worry about it if it would take a lot of work to avoid - when comparing changes I tend to put together little scripts that can delete the ID field so I ignore changes to it.

PR #488, opened 2026-02-24
Ultra Tech: Add Magazines for conventional guns

fishface60:
This PR does not structurally depend on https://github.com/richardwilkes/gcs_master_library/pull/495 but the ice gun magazine assumes the same values so logically depends on it.

This specifies Magazines according to the rules in Loadouts: Starship Crew.

Guns which are individually loaded, typically single-shot weapons and missile launchers such as the IML, TML, Gyroc Pistol and Holdout Gyroc do not have external magazines.

A number of magazines have zero weight because the total weight of the ammo equals the listed weight of the reload and this implausible result is a consequence of errors in Ultra-Tech.

Others additionally note "weight is erroneous" because the total weight of the ammo exceeds the listed weight of the reload and the numbers can't be made to add up.

### DIFF
--- a/Library/Ultra Tech/Ultra Tech Equipment.eqp
+++ b/Library/Ultra Tech/Ultra Tech Equipment.eqp
@@ -20613,7 +20613,7 @@
 				"Weaponry"
 			],
 			"base_value": "1400",
-			"base_weight": "1 lb",
+			"base_weight": "12 lb",
 			"weapons": [
 				{
 					"id": "WqtpFn_3ogNM9oiHv",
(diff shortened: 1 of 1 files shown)


richardwilkes:
So reading the text for Ice Gun (TL10), it states that "a cup of water" is used as ammo, turning the water into 20 ice needles. A cup of water is 8 oz, which is 0.5 lb (OK, technically those are fluid ounces, so it isn't quite that neat, but I'd leave it at this for nice values). Divide that by 20 and you get 0.025 per shot.

Not sure if that's the only question here, as I haven't looked through the actual file.

fishface60:
Ok, I'll resolve the erroneous magazines with Chandley's stats using the smallest magazine available to reduce deviation from UT RAW and record deviations with a justification.

Since the difference will be the reload weight and magazine weight this should have no impact to existing users of the equipment since the modifier and magazines are opt-in.

PR #497, opened 2026-03-03
Ultra Tech: Fix Gyroc LSW, 15mm weight

fishface60:


maw3193 (approved):
This all looks correct in the diff, in Ultra-Tech, and I've confirmed it looks right when loaded in gcs. approved.

PR #502, opened 2026-03-05
Add High-Tech Adventure Guns

maw3193:
Submitted by Best Boi on the gcs discord.

This is a mixture of existing weapons from High Tech Equipment, and new weapons from High-Tech Adventure Guns.

### DIFF
--- a/Library/High Tech/High Tech Adventure Guns.eqp
+++ b/Library/High Tech/High Tech Adventure Guns.eqp
@@ -1,0 +1,31079 @@
+{
+	"version": 5,
+	"rows": [
+		{
+			"id": "eH0aap2PykcJPZMSE",
+			"source": {
+				"library": "richardwilkes/gcs_master_library",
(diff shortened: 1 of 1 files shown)


rinickolous (changes requested):
Apart from the issues @fishface60 already pointed out and the unloaded weight modifiers these weapons could use, the primary thing I'm seeing straight away is the missing page references. A handful of entries carry the original High-Tech page references but the majority do not have a page reference at all. I'd like this amended. Given that the page location for some of these entries is non-obvious (e.g. the "Winchester Model 1882, .32-20 Winchester" is never listed, but its existence is hinted at through a mention of the Winchester being chambered in .32-20 Winchester from 1882), a page reference highlight where appropriate would be ideal (but is not strictly required for me to approve this).

rinickolous:
Superseded by #516 (which we'll merge as soon as the minor issues with that PR are corrected). Closing.

PR #505, opened 2026-03-05
Ultra Tech: Add modifier for ETC Guns

fishface60:
Following existing conventions there is a multiplicative cost version and a CF version.

Modifiers are provided for unloading the energy cell from the guns listed by category.


rinickolous (approved):

rinickolous (changes requested):
The Basic Set Multiplicative version of the modifier does not include the "Using unloaded cell stats" containers which the CF version does. Let's duplicate the two so users can find it either way

rinickolous (approved):

PR #514, opened 2026-03-18
High Tech: Add Unloaded modifiers to Enfield SA80

fishface60:
This adds the unloaded modifier for the L85A1, L85A2 and LMG variants.

They should use M4 30-shot magazines.

### DIFF
--- a/Library/High Tech/High Tech Equipment.eqp
+++ b/Library/High Tech/High Tech Equipment.eqp
@@ -25367,6 +25367,15 @@
 					}
 				}
 			],
+			"modifiers": [
+				{
+					"id": "fV-q-yqztnWvwvWAi",
+					"name": "Using unloaded stats",
+					"cost": "-34",
+					"weight": "-1 lb",
+					"disabled": true
+				}
+			],
 			"quantity": 1,
 			"calc": {
 				"value": 1300,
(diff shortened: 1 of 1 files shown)


maw3193 (dismissed):
I've confirmed that the weights are correct, and that it loads in gcs and looks right.

I dug into High-Tech and confirmed that they use the M16/M4 magazines (High-Tech says to use the M16 Magazine, though the M16A1 uses a different calibre, the M16A2 is 5.56x45mm NATO, just like the M4).

I think it's not ideal that it's hard to guess which magazine to use for unloaded stats (they can read it in High-Tech or guess from the calibre), but I don't think that blocks merging.

I've approved because I think this is mergeable, but will refrain from pressing the button in case you want to make changes to address which magazine to use.

maw3193 (changes requested):

rinickolous:
@fishface60, haven't heard from you on here for a few months. can we get these last little bits corrected so we can merge this?

maw3193:
I've spoken to @fishface60 privately and he doesn't have time to work on this but is happy for me to adopt these changes.

maw3193:
I think this can be closed, since its changes were included in #579 which was just merged.

PR #520, opened 2026-03-27
Add Forgeling to DFRPG Races

daliin:
Added the Forgeling racial template from the sidebar in DFRPG Monsters 2, p. 19.


daliin:
I'm amazed that you caught all of that! My second commit should address all the issues except the ones that require updating the DFRPG Traits library. (I'll submit another PR for that when I get a chance.)

rinickolous (changes requested):
Sorry for the wait here. This all looks fine now, except for the **Weakness** Trait. The book lists this as 1 HP per minute, whereas the Trait in the Master Library does not have this as an option. I would suggest adding a custom modifier (1 damage per minute) with -10 cost adjustment (it's basically the equivalent of 1d per 5 minutes), and changing "immersion in water" to be a replacement string for the "@\Occasional Substance@" modifier (since immersion in water is likely occasional rather than rare, and it'd even out the point cost).

daliin:
I think my latest commit adjusts things properly. Let me know if I should do anything else. Thank you!

rinickolous (approved):
No further issues. Merging.

PR #529, opened 2026-05-04
Fix Urban Jungle Gym

maw3193:
As reported on Discord by Jeffrywith1e, the Urban Jungle Gym trait was erroneously requiring a trait named Jumping at level 16, rather than the skill Jumping at level 16.

This fixes it in the Action Traits library, and both the templates that use it.


dripton (approved):
LGTM

PR #534, opened 2026-05-11
Add generic perks and quirks

maw3193:
As requested in #533.

Perks and quirks are less strictly-defined than other traits and can be whatever the GM and Players agree is worth 1 point.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -98,7 +98,8 @@
 		{
 			"id": "trPM4mpwMJnj-tqdg",
 			"name": "@Mental Quirk@",
-			"reference": "B162",
+			"reference": "B162, PU6:17",
+			"reference_highlight": "Mental Quirks",
 			"tags": [
 				"Mental",
 				"Quirk"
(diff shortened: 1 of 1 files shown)


rinickolous (changes requested):
`Basic Set Traits` already includes "@Mental Quirk@" and "@Physical Quirk@" which duplicates the scope of "Quirk: @Description@". Both of these existing items are used at least once elsewhere in the library. To not duplicate scope for these, either change the other master library instances to use the new Quirk, or remove the generic "Quirk: @Description" and replace the generic Quirk with a Physical and Mental one (with relevant tags) to stick to the existing format.

maw3193:
I've reworked this in line with the `@mental perk@` and `@physical perk@`. I've taken the liberty of creating Mental Perk, Physical Perk, Social Perk, and Supernatural Perk, and their equivalent quirks, and included page references for the Power-Ups books.

Basic Set only includes Mental and Physical quirks, but trait categories differentiate between Mental, Physical, Social and Supernatural traits (and what kind they are is relevant for situations like body swapping and possession).

rinickolous (approved):

PR #539, opened 2026-05-17
Basic Set: Add Trait and Equipment modifiers to add reach based on the wielder's SM

maw3193:
These trait and equipment modifiers for SM+1 to SM+10 that increase maximum reach, as described in _Size Modifier and Reach_ (see B402).

With thanks to Ptolemy for the original trait modifiers.

I used their trait modifiers to create equivalent equipment modifiers, and added some scripted notes to warn if they're being used incorrectly.

### DIFF
--- a/Library/Basic Set/Basic Set Enhancement Modifiers.adm
+++ b/Library/Basic Set/Basic Set Enhancement Modifiers.adm
@@ -2,6 +2,209 @@
 	"version": 5,
 	"rows": [
 		{
+			"id": "Mqb9tJcYmw5RpDY0d",
+			"name": "Size Modifier and Reach modifiers",
+			"reference": "BX402",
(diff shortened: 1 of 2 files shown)


maw3193:
I've discovered a bug with some of the scripted notes when applied to a trait that isn't levelled

maw3193:
I created https://github.com/richardwilkes/gcs/pull/1045 to fix the bug, but in the meantime I have removed the functionality that relied on it, so this is ready for review now.

maw3193:
I think it would also be a problem for the scripted checks to not take into account that Strikers can be Long and Extra Arm / Modified Arms can be Long or Short.

Rather than take into account every kind of complexity, I'll remove the size modifier check.

rinickolous (approved):
The whole note including the error sits inside of parentheses but I guess there's no way around it for now so that's fine.

PR #542, opened 2026-05-29
Add GURPS Furries

maw3193:
As requested by El Dourado on the GCS Discord

### DIFF
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -11189,6 +11189,24 @@
 			"points": 2
 		},
 		{
+			"id": "q6mA8uvi7Ixh57qv6",
+			"name": "Kicking",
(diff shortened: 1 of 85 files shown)


maw3193:
I'm taking this out of draft because it's now feature-complete.

This commit series includes supporting commits that changes non-furries parts of the master library, which have their own pull requests which should be merged first:
#543, #544, #546, #547, #548, #550, #552, #553, #554, #555, #556, #557, #558	

Strictly-speaking, I could remove those commits from this pull request and then they could be reviewed and merged independently, but the source library references may be a little confused.

maw3193:
I pushed changes to use the fixed Striking ST in all the templates

maw3193:
The latest pushed changes incorporates the fixes to Metabolism Control, and the fix to Uncongenial.

maw3193:
This now also uses #585 (Alternate Form)

PR #543, opened 2026-05-31
Basic Set Traits: Fix Striking ST's One Attack Only modifier's price

maw3193:
Previously, it was applying -60% to the price _per level_ of the advantage, whereas it should be -60% irrespective of trait level.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -24797,31 +24797,39 @@
 			],
 			"modifiers": [
 				{
-					"id": "mCVSG-r2Udy4p-_4W",
+					"id": "mrwY7GUU_azOnL6Um",
 					"name": "No Fine Manipulators",
 					"cost_adj": "-40%",
 					"disabled": true
 				},
 				{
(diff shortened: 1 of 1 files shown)


maw3193:
On second thought, I think this change is incomplete.

Striking ST's One Attack Only modifier must use level from owner, because its bonus is per level of this modifier.

However, the cost is calculated per level of this modifier, so One Attack Only (-60%) 3 would become a -180% cost modifier.

I think changing the price modifier from -60% to -3 satisfies the bonus being correct, and the price being correct for levels other than 1, but not being a percentile modifier means the price is calculated incorrectly when combined with other modifiers -
i.e. Striking ST 2 (one attack, -60%; size 2 -20%) is 2 points
Striking ST 2 (one attack, -3, size 2 -20%) is `ceiling(4 * 0.8)` = 4.

I think the only way to actually make this work in the current data model is to separate the modifier that sets the price from the modifier that applies the mechanical effect - e.g.
image

maw3193:
I've pushed changes in line with the previous comment.

rinickolous (approved):
Makes sense given the current limitations, though I would raise an issue about more intuitive support if one hasn't been raised already.

PR #544, opened 2026-05-31
Basic Set Traits: add Hot and Cold modifiers to Temperature Tolerance

maw3193:
In GURPS Furries, templates with thick fur have "Temperature Tolerance (Cold)". This adds modifiers for "Hot" and "Cold" (and implicitly the wider temperature range is evenly spread between hot and cold if neither modifier is picked).

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -32318,15 +32318,60 @@
 			"id": "touwUIFalnae02BUl",
 			"name": "Temperature Tolerance",
 			"reference": "B93",
+			"local_notes": "<script>\nif (entity.exists) {\n  var degreesHotter = 0;\n  var degreesColder = 0;\n  const hotMod = self.findActiveModifier(\"Hot\");\n  const coldMod = self.findActiveModifier(\"Cold\");\n  const degreesHotterMod = self.findActiveModifier(\"Degrees Hotter\");\n  const degreesColderMod = self.findActiveModifier(\"Degrees Colder\");\n  if (hotMod) {\n    degreesHotter = $ht * self.level;\n  }\n  if (coldMod) {\n    degreesColder = $ht *self.level;\n  } \n  if (degreesHotterMod) {\n    degreesHotter = degreesHotterMod.level;\n  }\n  if (degreesColderMod) {\n    degreesColder = degreesColderMod.level;\n  }\n  if (degreesHotter == 0 && degreesColder == 0) {\n    degreesHotter = $ht *self.level / 2;\n    degreesColder = $ht *self.level / 2;\n  }\n  var strings = [];\n  if (degreesHotter != 0) {\n    strings.push(degreesHotter + \"° hotter\");\n  }\n  if (degreesColder != 0) {\n    strings.push(degreesColder +\"° colder\");;\n  }\n  strings.join(\" and \")\n} else {\n  \"Will calculate how much the comfort zone is expanded when added to a character sheet\"\n}\n</script>",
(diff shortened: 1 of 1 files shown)


rinickolous (commented):
Is there a page reference for this anywhere that you could add?

maw3193:
I'm not sure if there's an explicit page reference for saying "hot" or "cold". B93 (the Temperature Tolerance trait) says `Each level of Temperature Tolerance adds HT degrees to your comfort zone, distributed in any way you wish between the “cold” and “hot” ends of the zone.`, so strictly-speaking each level of Temperature Tolerance gives you HT degrees to distribute, which could be implemented as levelled trait modifiers for "Hot" and "Cold" to set the edges of their comfort zone, which can be enforced with a scripted prerequisite once it's on a character sheet.

Realistically, I don't think it's ever used like that - it's either assumed to be evenly-distributed between hot and cold, all hot, or all cold. For example, Template Toolkit 2: Races gives Ravens `Temperature Tolerance 1 (Colder)`; Furries gives lots of the racial templates `Temperature Tolerance 1 (Cold)`, from thick fur or feathers; and Bio-Tech gives Selkies `Temperature Tolerance 1 (Cold)` and Tritons `Temperature Tolerance 2`.

maw3193:
I could add a page reference with highlight to the section about choosing how to distribute those degrees, but I think that may be confusing as the trait modifier doesn't actually do that.

I think the best I could do is:
* A trait modifier container named "choose how to distribute the degrees around your comfort zone" and referencing the page reference in my previous comment.
  * "Cold" distributing all degrees to the cold end.
  * "Hot" distributing all degrees to the hot end.
  * "Degrees Colder (trait modifier level)" degrees distributed to the cold end.
  * "Degrees Hotter (trait modifier level)" degrees distributed to the hot end.
  * Assuming the lack of any modifier means they're evenly distributed to the hot and cold end.
* Using scripting in the notes to describe how your comfortable temperature range has been extended
* Scripted prerequisites to check the trait modifiers have been used correctly
  * Don't mix Cold, Hot and Hotter/Colder.
  * If using Hotter/Colder, that the total levels of Degrees Hotter and Degrees Colder add up to HT * trait level.

maw3193:
I've updated this as described in the previous comment

maw3193:
Latest push was because I found a bug - failed prerequisite if no modifier was picked at all.

maw3193:
I rebased this and made sure there were no more automatic file format changes with the latest gcs versions.

Library/Basic Set/Basic Set Traits.adq
```diff
@@ -32318,15 +32318,60 @@
...
+					}
+				]
+			},
+			"modifiers": [
+				{
+					"id": "MI1VcdSeblizJBfA0",
+					"name": "Choose how to distribute degrees to the character's comfort zone",
+					"reference": "B93",
+					"reference_highlight": "Each level of TemperatureTolerance adds HT degrees to yourcomfort zone, distributed in any wayyou wish between the “cold” and “hot”ends of the zone.",
```
dripton: Several missing spaces

dripton (commented):
Looks reasonable.  Some missing spaces in description text.

Library/Basic Set/Basic Set Traits.adq
maw3193: Will sort - they were originally needed for the highlight across a multi-line page reference, but spaces works correctly with the updated PDF reader, and line breaks aren't the same between Basic Set and Basic Set Revised anyway.

dripton (approved):
LGTM now

PR #545, opened 2026-05-31
Basic Set Traits: add Long Neck

maw3193:
Technically, Long Neck comes from Furries, not Basic Set, but adding a modifier to the Basic Set trait helps keep all the modifiers in one place.


dripton (approved):
LGTM

PR #551, opened 2026-06-01
Power Ups: Add Easily Suppressed Disadvantage and its examples

maw3193:
Making a disadvantage easily-suppressed is a common way of turning one into a quirk.

This adds a generic easily-suppressed disadvantage with substitutions, and all the listed examples of Easily Suppressed Disadvantages.

### DIFF
--- a/Library/Power Ups/Power Ups Traits.adq
+++ b/Library/Power Ups/Power Ups Traits.adq
@@ -95,6 +95,20 @@
 			}
 		},
 		{
+			"id": "tG9VigVLJ8-NkRqiB",
+			"name": "Vainglorious",
+			"reference": "PU6:10",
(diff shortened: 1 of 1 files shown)


Library/Power Ups/Power Ups Traits.adq
```diff
@@ -12394,6 +12450,35 @@
 				"points": 1
 			}
 		},
+		{
+			"id": "tWGgEWs-7CasZce1b",
+			"name": "Swears on their father's grave",
```
dripton: I don't think we need the exact examples here.

dripton (commented):
This looks fine, but I think the exact examples of Compulsive Vowing are not needed.

dripton (approved):
LGTM

PR #553, opened 2026-06-02
Template Toolkit Races: Allow Potential Form to set its point value with levels

maw3193:
This is handy for a Potential Form that's worth more points - it has a point value equal to half the points of the new form, like Heir.


dripton (approved):
I don't have this book, but this looks simple and fine.

PR #559, opened 2026-06-17
Fix metabolism control mastery

maw3193:
As reported by Jackbauer on the gcs discord, Metabolism Control's "Mastery" enhancement was misspelled as "Maestry".
This fixes that typo everywhere in the master library (and incidentally adds source library information to various parts where it was already missing).


dripton (approved):
LGTM

PR #560, opened 2026-06-17
Add fantasy folk reptilians

maw3193:
This adds the racial templates from Fantasy Folk: The Reptilian Races.

Because this includes extensions to racial templates that were previously described in Banestorm, Dragons, or Dungeon Fantasy, I have created/updated the templates in their original place, and included a README.md for the Reptilian Races folder that includes markdown links to those templates.


maw3193:
So far, this depends on #546 and #547

maw3193:
I've finished work on this so I'm taking it out of Draft. It still depends on #544 and #585.

PR #563, opened 2026-06-25
Fix uncongenial

maw3193:
As mentioned in the gcs discord by El Dourado, Uncongenial's page reference is incorrect.

This does three things:
* Change the page reference for Basic Set Traits' Uncongenial trait to the correct one.
* Merge Basic Set Traits and Power Ups Traits' Uncongenial together by adding all the tags and page references from Power Ups' trait to Basic Set's, and then add a source library reference from Power Ups Traits' trait to Basic Set's trait.
  This means that only the Basic Set trait needs to be changed in future for changes to be picked up by source library sync.
* Go through all the sheets and templates in the master library and make sure they use the fixed Uncongenial trait.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -28323,7 +28323,7 @@
 		{
 			"id": "tVKcKQGhh9NHMQOvi",
 			"name": "Uncongenial",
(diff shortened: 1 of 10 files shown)


dripton (approved):
LGTM

PR #564, opened 2026-06-25
Basic Set Traits: Fix Vow's page reference

maw3193:
As reported in the gcs discord by El Dourado, Vow's page reference doesn't point to the page where Vow starts being described.

This fixes it in Basic Set Traits.

It also fixes the page reference to Basic Set for the vows in Action Traits.


dripton (approved):
LGTM

PR #566, opened 2026-06-28
Make Wealth and Status consistent in all the different trait lists

maw3193:
It was mentioned in the gcs discord that there was no Average Wealth trait or Status 0 trait.

I looked through all the places Wealth and Status are mentioned in the master library.

For status, there is a mix of:
1. Status as an advantage and Status as a disadvantage (slightly confusing, inconvenient when giving status due to wealth as a feature).
2. Status as two traits, High Status and Low Status (slightly confusing to have a trait named "High Status 0" to describe having Status 0).
To have a consistent treatment of Status that allows Status 0 to be described, I decided to have two traits, "Status" (default level is 0) and "Low Status".

For Wealth, there is a mix of:
1. A trait called Wealth with modifiers for the different wealth levels.
2. Multiple traits for each different level of Wealth.

Because traits with trait modifiers can be inconvenient for syncing (fields in the trait get synced but the modifiers stay unchanged), I decided to choose multiple traits.

Note that GURPS Lite doesn't have Multimillionaire, and DFRPG (without Dungeon Fantasy Companion 2) doesn't have Multimillionaire or Filthy Rich, so they are sometimes omitted from the list.

As a general plan to make it easier to keep things updated in the master library, I've merged these traits together in various trait lists - Basic Set Traits includes page references for Lite, Discworld Roleplaying Game and DFRPG, and their counterparts in Lite, Discworld Roleplaying Game and (but not DFRPG due to later changes) use the traits from Basic Set Traits, with source library information pointing back - With this merged, if Wealth or Status need to be changed again, they only *have* to be changed in Basic Set Traits. When those traits from other libraries get added to a character sheet, gcs will note that it differs from Basic Set Traits and can be synchronized.

Other, related changes are:
* Implemented the optional Wealth and Status rule, so you can enable a trait modifier to get a Status bonus at high levels of Wealth.
  * This includes a scripted prerequisite for Multimillionaire to make sure you select the right trait modifier
* Implemented Partial Multimillionaire Levels from Spaceships 2
  * This includes calculating how much a multiplier to starting wealth you get from your level of Multimillionaire, which can be a decimal.
  * It also includes altering the code for the Wealth and Status prerequisite to continue to behave correctly when the trait's level can be a decimal instead of an integer.
* It adds the higher wealth levels from Dungeon Fantasy Companion 2 to the Dungeon Fantasy RPG Traits list.
  * Because this is incompatible with keeping source library syncing with Basic Set Traits, I deleted the source library information and made them DFRPG-specific.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -8855,6 +8855,25 @@
 					}
 				]
 			},
+			"modifiers": [
+				{
+					"id": "mxxXX0DiOkPBXV89H",
+					"name": "Wealth grants Status",
+					"reference": "B26",
+					"reference_highlight": "Wealth and Status",
+					"features": [
+						{
+							"type": "trait_bonus",
+							"name": {
+								"compare": "is",
+								"qualifier": "Status"
+							},
+							"amount": 1
+						}
+					],
+					"disabled": true
+				}
+			],
 			"base_points": 50,
 			"calc": {
 				"points": 50
@@ -18147,6 +18166,10 @@
 				"all": true,
 				"prereqs": [
 					{
+						"type": "script_prereq",
+						"script": "const statusModifier = self.findActiveModifier(\"Wealth grants Status\");\nif (statusModifier \u0026\u0026 statusModifier.level == 2 \u0026\u0026 self.level != 1) {\n  \"Wealth grants Status 2 must be used with Multimillionaire 1\"\n} else if (statusModifier \u0026\u0026 statusModifier.level == 3 \u0026\u0026 self.level == 1) {\n  \"Wealth grants Status 3 must be used with Multimillionaire 2+\"\n} else if (self.level == 0) {\n  \"Multimillionaire should not be level 0\"\n} else { \n  \"\"\n}"
+					},
+					{
 						"type": "trait_prereq",
 						"has": false,
 						"name": {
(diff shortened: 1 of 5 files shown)


Library/Basic Set/Basic Set Traits.adq
```diff
@@ -18025,8 +18153,8 @@
 		{
 			"id": "tc2Bw-0JpkS5K2zxY",
 			"name": "Multimillionaire",
-			"reference": "B25",
-			"local_notes": "Starting wealth is \u003cscript\u003eformatNum(100 * Math.pow(10, self.level), true, false)\u003c/script\u003ex normal",
+			"reference": "B25, DFRC2:49, DW33",
+			"local_notes": "Starting wealth is \u003cscript\u003e\n// Partial Multimillionaire Levels (SS2:26) allows Multimillionaire level to be any multiple of 0.2.\nlet whole = Math.trunc(self.level), partial = Math.round((self.level - whole)* 10) || 1;\nformatNum(10 ** ( whole + 2) * partial, true, false)\n\u003c/script\u003ex normal",
```
dripton, line 18157: This is hairy and probably won't be used in many campaigns but I guess it's reasonable to include, since people who don't want it don't have to use it.

dripton (approved):
LGTM.  The scripting around partial multimillionaire levels is a bit complicated but nobody has to use it, so I guess it's better to have than not.

fishface60:
Not something you've done, but something I've noticed since you did work adjacent - the trait name for low status is `"Low Status "`, note the trailing space.

Source library information will make cleaning up later easier, so it doesn't need to be corrected now, but it would be nice to have if you make another pass over the changes.

maw3193:
> Not something you've done, but something I've noticed since you did work adjacent - the trait name for low status is `"Low Status "`, note the trailing space.
> 
> Source library information will make cleaning up later easier, so it doesn't need to be corrected now, but it would be nice to have if you make another pass over the changes.

Good catch! I'll take a look at that for a subsequent Pull Request.

PR #567, opened 2026-06-29
Fix the trailing space in Low Status' name

maw3193:
This fixes the Low Status trait in the Basic Set Traits, Low Tech Traits and Discworld Roleplaying Game trait lists.

It also updates all sheets and templates that use Low Status to use the fixed version.


dripton (approved):
LGTM

PR #571, opened 2026-07-03
Submission to Add a Homebrew folder and file

Tuvulu123:
Please review this .adq file. it is a set of Racial Packages for my fantasy campaign called "Orloth." Rich suggested, that if it met your approval, it should be placed in an "Orloth Custon" folder in Homebrew.


maw3193:
Since this is homebrew there's no standard for me to compare these to and say what's right or wrong, but I'll have a look and see if some things look odd to me.

I can't comment inline because the file's too big for github, so I'll comment here:
* Your Aquatic Lens has Enhanced Move (Water) with the Newtonian enhancement, which is usually meant for space movement, and I can't imagine a form of aquatic movement with a finite amount of thrust before having to refill on reaction mass.
* The Brunak has Code of Honor (Hearth Oaths) but there's no definition of what that is.
* The Dhampir has Doesn't Breathe with a custom modifier of "breathes half normal rate, -50%", but -50% is enough to hold breath for 25x as long as normal. I suspect that being able to hold breath for only twice as long would be at least -80% and may just be a perk (unless I've misunderstood how a slower breathing rate differs from longer breath holding).
* The Dhampir has a Vulnerability to Holy Water. I think by itself Holy Water does no damage, so this might a Weakness instead, or be paired with Divine Curse (Treat Holy Water as Acid) so that Holy Water becomes damaging.
* The Dragonborn has Resistant (Fire), giving a bonus to HT rolls against fire. I believe Fire typically harms as direct damage, so you'd want Damage Resistance (Limited, Fire, -20%).
* The Goldenbaugh Elf's Forest Affinity lists its effect in the notes, but it could be a feature to give that reaction modifier.
* Your Half-elf's Cultural Familiarity has the Alien enhancement - I think this would mean your elves are stranger than the average fantasy setting usually makes them.
* Your Half-orc has a Grim Visage perk which gives a bonus to intimidate - wouldn't this be a Racial Skill Bonus to the Intimidation skill worth 2 points instead?
* Halfling Fortune lists its point value in the notes. I'm assuming behind the scenes this functions as Luck (critical failures only, one reroll only, must be taken even if worse, -80%), which seems like a fair price to me.
* Your Tabaxi have Light Sleeper, so they have difficulty getting to sleep. I suspect this might be unintentional, as in GURPS Furries the cat people have the exact opposite trait, the Deep Sleeper perk. They may also/instead have Sleepy, meaning they need more hours of sleep per day.
* Your Tiefling's racial perk costs 2 points. I think Accessory (Lighter) would just be 1 point.
* For Wrought Ones, I think Social Stigma (Artificial Being) using both Obvious from your appearance and Obvious from your speech might have the unintended effect of stacking both reaction penalties together when you're seen and speaking at the same time (my mistake, when I made Social Stigma do this I hadn't considered that someone might use multiple modifiers).

maw3193:
Hi @Tuvulu123 , this pull request has been open for a while after I left some comments. Are you still interested in getting this merged?

PR #573, opened 2026-07-04
Add support for Ritual Magic from Basic Set and Magic

maw3193:
Context
===
The Ritual Magic system from Basic Set p.242 and Magic p.200, has one core skill, one skill per college, and one technique per skill.

It was raised in https://github.com/richardwilkes/gcs/issues/810 that defaults misbehave when there's a mixture of skills with specializations and skills without specialization.

I had a look at the problem, and it seems that Ritual Magic needs a separate skill for the core Ritual Magic skill, and a skill for each college that defaults to the core ritual magic skill, otherwise the Ritual Magic (Hermetic) skill could attempt to default to the Ritual Magic (Communication & Empathy) skill.

As of gcs v5.43.0, there is a context menu option to change the order that traits default to each other, but I think there could still be confusion over what the player needs to do.

Changes
===

* I created a new skill, Ritual Magic College (@College@), which defaults to a core skill and has prerequisites set up to warn if it exceeds its core skill.
  - The skill uses substitutions for the core skill (e.g. it could be Thaumatology instead of Ritual Magic); tradition (e.g. Hermetic, Voodoo or Witchcraft), since Ritual Magic specializes by tradition; and each individual college.
* I changed all of the ritual magic spells from the Ritual Magic Spells list to be based off "Ritual Magic College" instead of "Ritual Magic".
* I updated the Ritual Magery trait in Basic Set Traits to give a bonus to the Ritual Magic College skill too.
* I updated the Basic Set example character, William Headley, to use the new Ritual Magic College skills and the updated Ritual Magery and Ritual Magic Spells.

Potential issues
===

* I may be ignorant of some of the technicalities that led to Ritual Magic college skills using the Ritual Magic skill in the first place.
* The prerequisite counts for some of the Ritual Magic spells on William Headley's sheet differs from the counts listed in the spell table of GURPS Magic. I don't know if this is an error in gcs or errata that I can't find.
* Perhaps I should create a version of Ritual Magic College for every college, in addition to having the substitution.
* I have not checked over the entire master library whether there are other uses or implementations of Ritual Magic that should also be updated.

### DIFF
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -21804,6 +21804,56 @@
 			"points": 1
 		},
 		{
+			"id": "sI_NfZOq4jCXMBCTM",
+			"name": "Ritual Magic College",
+			"reference": "B242,  M200",
+			"reference_highlight": "Ritual Magic",
+			"local_notes": "AKA \"College of @College@\" or \"Path of @College@\"",
+			"tags": [
+				"Magical",
+				"Occult"
+			],
+			"specialization": "@College@",
+			"difficulty": "iq/vh",
+			"defaults": [
+				{
+					"type": "skill",
+					"name": {
+						"compare": "is",
+						"qualifier": "@Core Skill@"
+					},
+					"specialization": {
+						"compare": "is",
+						"qualifier": "@Specialty@"
+					},
+					"modifier": -6
+				}
+			],
+			"prereqs": {
+				"type": "prereq_list",
+				"all": true,
+				"prereqs": [
+					{
+						"type": "skill_prereq",
+						"has": true,
+						"name": {
+							"compare": "is",
+							"qualifier": "@Core Skill@"
+						},
+						"specialization": {
+							"compare": "is",
+							"qualifier": "@Specialty@"
+						}
+					},
+					{
+						"type": "script_prereq",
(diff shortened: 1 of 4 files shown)


rinickolous:
@maw3193 could you fix the merge conflicts please?

maw3193:
rebased, merge conflicts should be fixed now.

Library/Basic Set/Basic Set Skills.skl
```diff
@@ -21800,6 +21800,56 @@
 			],
 			"points": 1
 		},
+		{
+			"id": "sI_NfZOq4jCXMBCTM",
+			"name": "Ritual Magic College",
+			"reference": "B242,  M200",
+			"reference_highlight": "Ritual Magic",
+			"local_notes": "AKA\"College of @College@\" or \"Path of @College@\"",
```
rinickolous: Needs space between AKA and quotation mark. No other issues as far as I can see.

rinickolous (changes requested):

Library/Basic Set/Basic Set Skills.skl
maw3193: sorted!

maw3193:
I rebased this against main and made sure that any automated file format changes from recent changes to the gcs application were included too.

This has grown a couple of commits, but the changes should still be readable commit-by-commit, and the automated file format commits can be easily dropped if another pull request fixes them first.

maw3193:
I've rebased against the main branch and converted the files I change to v5.50.0 (in separate commits)

PR #574, opened 2026-07-05
Add updated optional specialties

maw3193:
This makes sure that every skill listed in Basic Set that calls out that it may have optional specialties has their optional specialties set up using the new Optional Specialization field.

Unfortunately, due to my lack of foresight the diff is a mess, as I haven't split the automatic file format changes to Basic Set Skills from the deliberate changes.

For completeness, the full list of skills affected is:
* Artist
* Astronomy
* Biology
* Cryptography
* Chemistry (no explicit optional specialty in the list, but it's the example for optional specialties)
* Intelligence Analysis
* Merchant
* Physics
* Veterinary

### DIFF
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -1260,6 +1260,29 @@
 			"points": 1
 		},
 		{
+			"id": "sfK9axZ_b-gCStIjJ",
+			"name": "Artist",
+			"reference": "B179",
+			"tags": [
+				"Arts",
+				"Entertainment"
+			],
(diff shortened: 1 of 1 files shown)


maw3193:
This is based on the work of #526. If this pull request is merged then #526 should be closed.

Fazzleburt:
Cooking also has Optional Specialties. Literature, Mathematics (Pure), and Surgery are mentioned for Optional Specializations but not given explicit examples.

maw3193:
Good catch, thanks! Cooking, Literature and Pure Mathematics updated, optional specializations added to Surgery

rinickolous (approved):

PR #575, opened 2026-07-05
Fix point costs for traits in the Inhuman Vampire template

TELunus:
I noticed that the Inhuman Vampire template was showing as a 194 point template even though it had no choices, but Monster Hunter expects the Inhuman racial templates to be 200 points each. Comparing the point costs of the traits against the list on MH1-51 I noticed "Can burn HP for extra effort" was 0 points on the template vs. 1 point in the book, and Draining (blood) was -10 points on the template vs. -5 points in the book. I've fixed both of these issues, bringing the template costs up to the correct 200 points.

I made the changes in GCS itself. It looks like GCS also decided to update some IDs and switch away from a Unicode escape for the ampersand in "Acute Taste & Smell"

### DIFF
--- a/Library/Monster Hunters/Classes/Champions/Inhuman Vampire.gct
+++ b/Library/Monster Hunters/Classes/Champions/Inhuman Vampire.gct
@@ -12,7 +12,7 @@
 					"children": [
 						{
 							"id": "tsWvCW1Guqeebb4FS",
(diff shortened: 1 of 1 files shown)


maw3193 (approved):
Thanks! This all looks correct - approved!

PR #578, opened 2026-07-12
DFRPG Traits: move perks, quirks and Wealth traits into trait containers

maw3193:
As requested by ArchBeth on the gcs discord, traits are organised in the list in a somewhat counter-intuitive way as a result of being sorted alphabetically and then being modified organically.
Putting Perks and Quirks in trait containers makes it easier to find them, and putting Wealth traits into a trait container named "Wealth" keeps them grouped together even if they might get sorted.


dripton (commented):
I am +1 on putting all the Wealth levels in trait containers.

I am +0 on quirks and perks.  I can see some benefits, but those are also huge lists of unrelated things, and you can already filter by quirks or perks.

I think this should probably be 3 commits and 3 pull requests rather than 1.

rinickolous:
I argee with @dripton on both counts. The tags should be sufficient for filtering to only Perks or Quirks. If the manual sorting of quirks is an issue then the solution is just to commit a change that sorts the file alphabetically or instruction to affected users to sort by name if they want to have an easier time finding something.

maw3193:
Okie doke, I'll strip out the perk and quirk changes then request review again, and create another pull request to sort all the traits alphabetically so that change can be considered separately.

maw3193:
The perk and quirk changes have been removed from this Pull Request

dripton (approved):
LGTM

PR #579, opened 2026-07-12
High Tech: Add Unloaded modifiers to Enfield SA80

maw3193:
This adds the unloaded modifier for the L85A1, L85A2 and L86A1 and L86A2 variants.

It also includes a dedicated magazine, identical to the Colt M16's, as HT118 notes that it uses detachable M16 Magazines.

This is a continuation of https://github.com/richardwilkes/gcs_master_library/pull/514


rinickolous (approved):
that automated change diff size tho 👀

PR #582, opened 2026-07-15
DFRPG Traits: sort the list alphabetically

maw3193:
Trait lists could be arranged manually, or sorted automatically.

Being arranged manually could be a result of adding traits in the order that they're listed in the book - this *could* be approximated by sorting by page reference, and then sorting by name where the page references are identical.

This can cause problems if you accidentally sort the list while it's being edited.
* If it wasn't being edited, you could just close the tab and discard the changes, preserving the manual arrangement.
* Because you're making changes, you now have to choose between destroying the original arrangement, trying to recreate it from memory, or discarding your changes and starting over.

I think it would be better for trait lists that the ordering isn't obvious and easy to fix, for the trait list to always be automatically sorted, and alphabetically A-Z is an easy one to scroll through to search for things in by eye.

These changes are built on top of the changes in #578 and should be merged after #578.

### DIFF
--- a/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Traits.adq
+++ b/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Traits.adq
@@ -40,6 +40,44 @@
 			}
 		},
 		{
+			"id": "tpqSHJrZERB3bRKz5",
+			"name": "@Mental Quirk@",
+			"reference": "DFA68",
+			"tags": [
+				"Mental",
+				"Quirk"
+			],
+			"base_points": -1,
+			"calc": {
+				"points": -1
+			}
+		},
+		{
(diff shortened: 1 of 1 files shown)


richardwilkes:
Agreed regarding the default to sorting alphabetically in ascending order. That's what the files used to do, but clearly some have diverged over time.

dripton:
There was a merge conflict around Forest Guardian.  I resolved it, but it might be worth another look before we merge this PR.

dripton:
I merged #578 which this was based on.

maw3193:
Thanks! I've rebased against main and checked Forest Guardian is behaving as it should - it should be ready to merge now

dripton (commented):
LGTM

dripton (approved):

PR #583, opened 2026-07-15
High Tech Equipment: Fix the MP5 and add missing variants

maw3193:
This does:
* Fix the cost of the MP5's magazines double-counting the cost of ammunition
* Fix the MP5's RoF being erroneously listed as 18 instead of 13.
* Add the MP5, MP5A1 to MP5A5 and MP5N variants.
* Rename the MP5's magazines to not be MP5A3-exclusive.
* Change the RoF of the MP5/10A3 and MP5/40A3 to have their highest RoF marked as full-automatic.

### DIFF
--- a/Library/High Tech/High Tech Equipment.eqp
+++ b/Library/High Tech/High Tech Equipment.eqp
@@ -46989,6 +46989,187 @@
 			}
 		},
 		{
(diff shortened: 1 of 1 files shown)


rinickolous (approved):
Looks good to me.

PR #584, opened 2026-07-16
Dungeon Fantasy 8: Add Decorative Embellishments

maw3193:
This includes all of the Decorative Embellishments from DF8 from page 53 to 55, except Jeweled, since the cost of jewels varies and could be managed by turning the clothing into an equipment container and adding jewels from DF8 Equipment.

This was contributed by Savar via Discord.

I have made some minor tweaks from Savar's contribution, including:
* Put them in a file named "Dungeon Fantasy 8 Equipment Modifiers", since there are other equipment modifiers in DF8 that could also be added to the Master Library.
* Change the tag from "Embellishment" to "Decorative Embellishment" to contrast with "Supernatural Embellishment"s if added later.
* Sort them alphabetically (manual ordering can get accidentally destroyed by automatically sorting and create untidy diffs that are harder to review).


rinickolous (approved):
No issues jump out at me.

PR #585, opened 2026-07-19
Basic Set Traits: Update alternate form to name the form and specify the cost

maw3193:
Alternate Form lets you shapeshift into one other form, at a base cost of 15 points, plus 90% of the difference in cost between your base racial template and the template of the alternate form (usually, some options remove the discount).

To support this without breaking the source library link, I've added a substitution value for the name of the other form, and allowed the cost of the form to be input using the level of the trait.


rinickolous (approved):
I _think_ that point cost with those modifiers looks about right. No obvious issues. Approved.

PR #590, opened 2026-07-30
DFRPG Traits: Add common kinds of Immunity as hard-coded traits

maw3193:
As contributed by ArchBeth on the GCS discord.

Some immunities are extremely common, and it's convenient to be able to drag the trait on without having to pick a modifier and name it every time.


Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Traits.adq
```diff
@@ -5557,6 +5557,57 @@
...
+			"id": "t1TH0fW2NUd-uk2MV",
+			"name": "Immunity",
+			"reference": "DFM11",
+			"local_notes": "Disease",
+			"tags": [
+				"Advantage",
+				"Physical"
+			],
+			"replacements": {
+				"Occasional: Disease, Ingested Poison, etc.": "Disease"
```
rinickolous: I don't think this `replacements` key is supposed to be there considering the corresponding @-sign encapsulated string doesn't appear anywhere. Must have not been pruned.

rinickolous (dismissed):

Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Traits.adq
maw3193: Indeed! There doesn't seem to be any way to remove it from the gcs gui, but everything works fine if I remove it manually.

richardwilkes (approved):

PR #591, opened 2026-08-01
Fix typo in Power-Ups talent

fdenzer:
Checked the book (PDF), it clearly is Packing, not a non-existent skill.


maw3193 (approved):
I think this change is correct. I can confirm that Beastmaster should have Packing and that's what this change seems to do. Approving now, though I'm away from desk right now so can't load it in GCS to check, otherwise I'd merge it now.

maw3193:
I've confirmed that this looks correct when pulled and run through GCS. looks good to me, thanks!

PR #594, opened 2026-08-14
Correct tech level of TL8 9mm Auto Pistol to 9

CaseyCarter:
As indicated on page 278 of the Basic Set Fourth Edition 7th printing PDF.


maw3193 (approved):
I've confirmed that the 9mm Auto Pistol should be TL9, and that this change makes it so.

PR #595, opened 2026-08-15
Druidic Animal Control spell names are incorrect

LeeSaferite:
In DFS17, the "(Animal) Control" group of spells are grouped under a single entry with each spell name being distinct only in the `Cost` section. The names used in the DFRPG Spells list were formatted differently (`Animal Control (Bird)` vs `Bird Control`). Additionally, there was a spell called `Animal Control (@Animal@)` that isn't an actual spell and shouldn't have been added.

On feedback from @maw3193 I moved the 5 animal category spells into an `Animal Control` spell container to keep them grouped and sorted under "A". This keeps the existing sorting behavior after the fix.


maw3193 (approved):
This looks correct to me - the changes are as described and the changed files look correct when loaded in GCS.

PR #596, opened 2026-08-15
Druidic Repel Animal spell names are incorrect

LeeSaferite:
In DFS17, the "Repel (Animal)" group of spells are grouped under a single entry with each spell name being distinct only in the Cost section. The spell calls out that it's costs are identical to the counterpart `(Animal) Control` spells. The names used in the DFRPG Spells list were formatted differently (`Repel Animal (Bird)` vs `Repel Bird`) than what they should be (based on how `(Animal) Control` names spells). Additionally, there was a spell called `Repel (@Animal@)` that isn't an actual spell and shouldn't have been added.

On feedback from @maw3193 I moved the 5 spells into a Repel Animal container to keep them grouped and sorted under "R". This keeps the existing sorting behavior after the fix.

### DIFF
--- a/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Spells.spl
+++ b/Library/Dungeon Fantasy RPG/Dungeon Fantasy RPG Spells.spl
@@ -9739,232 +9739,207 @@
 							"points": 1
 						},
 						{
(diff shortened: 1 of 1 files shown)


maw3193 (approved):
These changes look to be correct as described and look correct when loaded in GCS too.

PR #597, opened 2026-08-15
Banestorm: Freelance Wizard Template Errata

LeeSaferite:
I found a mistake in the published Banestorm Freelance Wizard template and reported it to SJG. They finally got back with the errata update that dropped 1 point in the `Innate Attack (Beam)` skill and added another `Air` spell (`Body of Air`) instead. Point balance is retained and the prereq of '6 other Air spells' is now met for `Lightning`.

### DIFF
--- a/Library/Banestorm/Classes/Freelance Wizard.gct
+++ b/Library/Banestorm/Classes/Freelance Wizard.gct
@@ -1447,7 +1447,7 @@
 													"modifier": -2
 												}
 											],
-											"points": 2
+											"points": 1
 										}
 									]
 								},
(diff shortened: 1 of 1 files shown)


LeeSaferite:
This errata is based on an email response I received from Nikki, the Errata Coordinator.

```
I received information about the correction to the freelance wizard template in GURPS Banestorm.

The Air and Earth spell list should be:

Air and Earth: Innate Attack (Beam) (E) DX [1]-11; and Body of Air, Create Air, Earth to Air, Lightning, Purify Air, Seek Earth, Shape Air, Shape Earth, and Walk On Air, all (H) [1]-15+.

We will make updates as soon as we can (though we don't have a timetable for errata updates).

Thank you again for pointing this out!

Nikki
Errata Coordinator
Steve Jackson Games
```

maw3193 (approved):
The changes look correct as-described.

I am concerned that Banestorm page 202 says `The spell-using templates (e.g., the Freelance Wizard and Priest-Wizard) use only spells from the Basic Set`, and that Body of Air isn't in Basic Set, but I've confirmed via Discord that's what's said in the amended spell list.
I consider actually meeting Lightning's prerequisites a significant improvement and making the statement on page 202 incorrect to be a minor drawback.

PR #599, opened 2026-08-20
Tactical Shooting: fix deadeye

### LINKED ISSUE #598: Deadeye (TS38) Should be levelled
The text for the Deadeye perk (TS38) says that you can take it up to three times, each point giving you a 10% reduction in the time it takes for Precision Aiming blah, blah, blah.

But gcs doesn't let you level it without editing the perk. It would be nice if it obeyed the text please.

Expected behaviour: (except the 2 in Levels)

Image

maw3193:
As reported in #598, The Deadeye trait should be a levelled trait with a maximum level of 3.

Because this trait is present in both Gun Fu and Tactical Shooting, I've updated it in both and made the Tactical Shooting one treat Gun Fu's as its source.

Closes #598

### DIFF
--- a/Library/Gun Fu/Gun Fu Traits.adq
+++ b/Library/Gun Fu/Gun Fu Traits.adq
@@ -197,11 +197,12 @@
 		{
 			"id": "tna6P7zNt41Qt3n4-",
 			"name": "Deadeye",
-			"reference": "GF18",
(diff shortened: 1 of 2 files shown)


richardwilkes (approved):

dripton (approved):
LGTM

PR #600, opened 2026-08-25
The Pressure Support trait needed a level limit and level-based note

LeeSaferite:


maw3193 (approved):
This change looks correct to me and I've confirmed it behaves as expected.

PR #602, opened 2026-08-25
Updates to the Basic Set "AI" Meta-Trait

LeeSaferite:
This is fixing broken references to the Basic Set traits, fixing the page ref search term, and adding the preconfigured flag.


maw3193 (approved):
These changes look correct and work as described. Approved!

PR #603, opened 2026-08-25
Updates to Basic Set "Body of Air" Meta-Trait

LeeSaferite:
- Added missing Basic Set source references
- Syncronized with Basic Set Traits
- Added prereq on "ST 0" trait to ensure ST stays at zero
- Added preconfigured flags
- Switched to use replacements instead of editing trait modifier name

### DIFF
--- a/Library/Basic Set/Meta-Traits/Body of Air.gct
+++ b/Library/Basic Set/Meta-Traits/Body of Air.gct
@@ -11,38 +11,55 @@
 			"container_type": "meta_trait",
 			"children": [
 				{
-					"id": "tv0ail4ONycImV1cO",
-					"name": "Attribute modifiers",
-					"local_notes": "ST 0 (may not be raised), HP +10",
-					"tags": [
-						"Physical"
-					],
-					"prereqs": {
-						"type": "prereq_list",
-						"all": true,
-						"prereqs": [
-							{
-								"type": "attribute_prereq",
-								"has": true,
-								"qualifier": {
-									"compare": "is"
-								},
-								"which": "st"
-							}
-						]
-					},
+					"id": "TGIq9E5uGIC0tthqW",
+					"name": "Attribute Modifiers",
 					"preconfigured": true,
-					"base_points": -80,
-					"features": [
+					"children": [
 						{
-							"type": "attribute_bonus",
-							"attribute": "st",
-							"amount": -10
+							"id": "tcxN7e51dSqhn29vd",
+							"name": "ST 0",
+							"prereqs": {
+								"type": "prereq_list",
+								"all": true,
+								"prereqs": [
+									{
+										"type": "attribute_prereq",
+										"has": true,
+										"qualifier": {
+											"compare": "is"
+										},
+										"which": "st"
+									}
+								]
+							},
+							"preconfigured": true,
+							"base_points": -100,
+							"features": [
+								{
+									"type": "attribute_bonus",
+									"attribute": "st",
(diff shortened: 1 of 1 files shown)


maw3193 (changes requested):
These changes all work but:
* Flight has no source library information - I suspect this was an oversight.
* I think the Attribute Modifiers trait could be decomposed into:
  - Decreased Strength 10
  - Extra Hit Points 10
  - Taboo Trait (Fixed ST)

There isn't a Taboo Trait (Fixed ST) in any of the trait lists that uses prerequisites to enforce a trait level.
The data model doesn't support setting up a generic `Taboo Trait(Fixed @Attribute@: @Attribute Level@)`, because the prerequisite statement "has the attribute ST which is <number>" doesn't allow the trait name or the number to be a substitutable text field. It is possible with scripting but that should ideally be used sparingly.

I think a Taboo Trait (Fixed ST) that was created for this template and enforced ST is 0 would be appropriate, though given a Fixed ST of 0 is common for various races and meta-traits, it would also be good as a new trait in Basic Set Traits.

Requested changes:
* Add source library information for Flight.
* Split the Attribute Modifiers trait into Decreased Strength, Extra Hit Points and Taboo Trait.
  - It would be nice, but not required, for Taboo Trait to come from a trait list as well.

LeeSaferite:
@maw3193 Ok, PR updated.

maw3193:
It looks like you have the attribute modifiers as:
* A trait container called Attribute Modifiers
* A trait called ST 0 which reduces ST by 10 and has a prerequisite that ST is 0.
* A trait called HP +10

I'd prefer an alternative:
* I think that you don't need a trait container for Attribute Modifiers.
* I think that you could use the existing Decreased Strength trait at level 10, and the Extra Hit Points trait at level 10.
* I think having the "ST 0" trait renamed "Taboo Traits (Fixed ST)", and not altering the ST attribute or changing the point value, would be better.

What do you think?

LeeSaferite:
@maw3193 The "Attribute Modifiers" group is a common pattern in a lot of templates. I don't mind removing it, but it's in common use. Additionally, the structure of my change was to convert that single custom "Attribute modifiers" trait into a group with the same name (with a fix on caps), then add individual traits to cover the parts of the original trait. If the preferred method to communicate the ST 0 restriction is via a leveled "Decrease Strength" (why isn't it called "Decrease ST"?) and then rename the custom restriction to "Taboo Traits (Fixed ST)", I'm fine with those changes. It might be nice to see if it makes sense to lift the taboo to basic traits instead of the template.

maw3193:
Hm, I wasn't aware that there were quite so many places that use a bespoke "Attribute Modifiers" trait - I count about 16, including Basic Set, Banestorm, Fantasy, and Dungeon Fantasy Adventure 1 (and one in Dungeon Fantasy 5 traits, from using Basic Set's Body of Air meta-trait).

There are a lot more places where templates set attribute modifiers by composing them out of individual traits, e.g. the Herakles racial template for Bio-Tech, or the Body of Air elemental meta-trait from Template Toolkit 2: Races.

Given there are a bunch of places that use Attribute Modifiers as a trait, I'm happy to be merged as-is, but would be glad to have the attribute modifiers composed of individual traits that come from Basic Set Traits.

I would be very glad to have "Taboo Traits (Fixed ST)" in Basic Set Traits, so that I can use it in future.

Let me know which approach you'd like to take - if it's to leave it as-is then I'll merge straight away.

PR #604, opened 2026-08-26
add basic set style html output template

Zuljita:
Adds a self-contained HTML export template using a classic two-page GURPS/DFRPG-inspired character-sheet layout.

Features include:

Fixed two-page core character sheet familiar to users of GCA or the GURPS Basic set character sheet
Automatic continuation pages for traits, skills, weapons, equipment, notes, and spells
Unified equipment rows for aligned item, state, cost, and weight data
Alternating gray and white rows for print legibility
Automatic “Page x of y” print numbering
Letter-size print layout with repeating continuation headings
The template was tested with GCS 5.48 using ordinary, spell-heavy, and deliberately overfilled characters. Test exports completed without unresolved template fields or clipped continuation rows.

### DIFF
--- a/Library/Output Templates/HTML - Basic Set Layout
+++ b/Library/Output Templates/HTML - Basic Set Layout
@@ -1,0 +1,1839 @@
+GCS HTML Template v1
+<!doctype html>
+<html lang="en">
+<head>
+	<meta charset="utf-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
+	<title>{{.Name}} - GURPS Character Sheet</title>
+	<style>
+		:root {
+			--ink: #231f20;
+			--paper: #ffffff;
+			--rule: #797979;
+			--stripe: #e8e8e8;
+			--serif: Georgia, "Times New Roman", serif;
+			--sans: Arial, Helvetica, sans-serif;
+		}
+
+		* { box-sizing: border-box; }
+		.template-source { display: none !important; }
+
+		html {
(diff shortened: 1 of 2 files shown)


richardwilkes:
I noticed the equipment section doesn't show quantity, which seems like something you'd want.

Also, conditional modifiers aren't on the sheet anywhere, either. Not essential, but having them listed does make life easier for a player.

One last thing: the name of the file should probably something like "HTML - PC (Book).html" or something similar, consistent with the existing output template files.

Zuljita (commented):
Fixed the missing pieces from @richardwilkes feedback and removed a little more dead space.

richardwilkes:
I few more things I noticed with the latest copy:

- Pagination (when printing or exporting to PDF) seems off a bit, as I'm getting extra blank pages between each real page.
- Basic Move is overlapping its value, while Basic Speed has lots of excess space.
- The word "Current" for HP overlaps the Appearance field above it.
- None of your weapon blocks list the skill level.

I've only quickly glanced over the actual text in the file so far, mainly focusing on the output for now.

richardwilkes:
I've reviewed the embedded javascript and found no security issues. I didn't consider things like efficiency or quality -- just that it didn't try to do things that would be bad for a user.

Zuljita (commented):
@richardwilkes thanks for taking the time to review previous iterations.  This should address your (accurate) concerns and be much better overall. Thanks!

PR #605, opened 2026-08-26
Add Defaults to Stage Combat

Enzarel:
Defaults per B222: Combat Art, Combat Sport... and a non-substitution placeholder for "actual combat skill"; as filtering skills by tags is not currently supported here.

A word of justification:

`@Substitution@` would ask the user once, upfront, for a specific skill to default to; I believe that's disingenuously simple. By the rules, the skill used here should change if another one is ever improved enough. Currently that'd be a manual process. Having players misunderstand that is not great, imo.

I trust hiding the line to be changed inside the skill editor is a good test for "are you experienced enough to understand how SC's defaulting works, and to swap out the placeholder for your highest skill as your character grows"

### DIFF
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -23404,6 +23404,30 @@
 					"type": "skill",
 					"name": {
 						"compare": "is",
+						"qualifier": "Combat Art"
+					},
+					"modifier": -2
+				},
+				{
+					"type": "skill",
+					"name": {
+						"compare": "is",
+						"qualifier": "Combat Sport"
+					},
+					"modifier": -2
+				},
+				{
+					"type": "skill",
+					"name": {
+						"compare": "is",
+						"qualifier": "[ Placeholder for actual combat skill ]"
+					},
+					"modifier": -3
+				},
+				{
+					"type": "skill",
+					"name": {
+						"compare": "is",
 						"qualifier": "Performance"
 					},
 					"modifier": -3
--- a/Library/Basic Set/Basic Set Skills.skl
+++ b/Library/Basic Set/Basic Set Skills.skl
@@ -23420,7 +23420,7 @@
 					"type": "skill",
 					"name": {
 						"compare": "is",
-						"qualifier": "[ Placeholder for actual combat skill ]"
+						"qualifier": "@Highest real combat skill@"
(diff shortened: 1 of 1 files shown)


maw3193 (dismissed):
I would be satisfied with a `@Highest Combat Skill@` substitution value, as I think I have the opposite opinion on discoverability - most users won't know that it's an option without opening up the skill and scrolling down.

On the other hand, I agree that it will cope poorly with some other skill becoming the highest combat skill, and I think having to fill in a substitution for a highest combat skill would be awkward for templates and non-combatant characters.

I'm happy to have this merged as-is, and I'll raise a feature request for skill defaults to search by tag

richardwilkes:
To do this right, each possible Combat Art/Sport should actually be created separately (no substitution logic), since the Basic Set states that they have the exact same defaults as the skill they are mimicking, but at an additional -3. That's a lot of work (and adds a lot of new entries), but is the only way to make this work correctly without users having to know when to fiddle with it.

maw3193:
Hm, and the recommended way of describing it is "<weapon skill> Art" and "<weapon skill> Sport", rather than a skill named "Combat Art" or "Combat Sport" with a specialization for the name of the weapon skill - which would be cumbersome to express by name for each of them.

I think this change, as it exists, is an improvement on the status quo, but there's still more work to be done, so I'm still happy for this to be merged as-is, and I'll create an Issue in the master library's project to describe the problem.

maw3193:
I've written up the issue at https://github.com/richardwilkes/gcs_master_library/issues/608.

I'm happy for this PR to be merged, but not certain that everyone else is happy for it to be merged now without further discussion.

Enzarel:
On second thought I agree with maw's first comment; substitution (`@Highest real combat skill@`) does this skill more justice right now. If there's a way to do it better in the future I'll revisit it either way. Pushed a commit, sorry for the extra button click.

As for Combat Art/Sport; reminder that Stage Combat doesn't specialize, and doesn't require Art/Sport to do so either. 😉 This is handled by the `known Name, any Specialty` feature perfectly, at -2. Only the worse "actual combat skill-3" is troublesome.

If this changes I'll be happy to fix it, though!

maw3193 (approved):
Agreed on no need to specialise, and glad to have your continued attention on this. Approving and merging.

PR #606, opened 2026-08-26
PY65: Minor update to Natural Weapon formatting.

Enzarel:
At some point between #485 and now, handling of line breaks in notes fields changed.

- Fixed spontaneous line breaks between:
  - the word "weight" and the weight value;
  - weapon quality and the weight line.
- While I'm down here... (There was extra space, so why not?)
  - Expanded "Wt." to the full word "weight"
  - Added "lb." unit.

Also 20 or so lines of automatic non-visible changes by GCS. Fun.

Screenshot of changes:
image


maw3193 (approved):
These all look correct and seem to behave correctly when loaded in gcs and applied to a character sheet. approved!

PR #609, opened 2026-08-26
Basic: Implemented Frequency of Appearance system from v5.41

Enzarel:
Removed Frequency modifiers and changed the Frequency of Appearance dropdown from `None Required` to `9 or less (Fairly often)`, for ×1 cost. Affects the following traits:
- Ally
- Patron
- Contact
- Dependent
- Enemy

Which, unless I missed something, is every trait that uses FoA in the Basic Set.

### DIFF
--- a/Library/Basic Set/Basic Set Traits.adq
+++ b/Library/Basic Set/Basic Set Traits.adq
@@ -1713,48 +1713,6 @@
 					]
 				},
 				{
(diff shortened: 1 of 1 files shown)


maw3193 (approved):
I've confirmed that these changes look correct and behave as expected. Approved, and merging.

PR #615, opened 2026-09-01
Updated scripting on High-Tech Scopes, utilizing new Substitutions.

Enzarel:
Version 2.0 of #463

Previously, the scripts here were comically bulky to galvanize against inputs outside the expected 0-5 range (including non-numbers); now that GCS supports limiting `@substitutions@` to a dropdown, most of this clunky safety bloat is no longer necessary.

In short, this PR changes the code from;

> `if (@Acc@ is a number, and is not a NaN, and is not something ridiculous, and please please please don't break the script) then (doMath) else (NukeEverything)`

to:

> `remember @Acc|1|2|3|4|5@ is a number, then doMath, then makeItPretty<3`

Regarding that last part, Variable-Power Scopes now display only set Acc values, hiding un-set ones. Works in any combination, and breaks _only slightly_ if none are set;

image

> Roll of Honor:
_@​LeeSaferite_ for making this possible with their amazing _richardwilkes/gcs​#1113_
_[Negative-Magnification Scope](https://github.com/user-attachments/assets/8512b88e-de65-4279-a211-6acc3408690f)_, brutally murdered but never forgotten ;(

### DIFF
--- a/Library/High Tech/High Tech Equipment.eqp
+++ b/Library/High Tech/High Tech Equipment.eqp
@@ -34415,60 +34415,69 @@
 			"id": "eyPjJiwnhI80yrneq",
 			"description": "Fixed-Power Scope",
 			"reference": "HT156",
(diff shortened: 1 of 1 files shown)


maw3193 (approved):
The code all looks correct - I was concerned from just looking at raw text about the existence of a `.filter(value => !isNaN(value) )`, but it's easier to read in the notes editor and see the context for it - you can have up to three different scope levels of accuracy and may leave one of them undefined.

I've also verified that it behaves as expected in GCS - approved!

PR #616, opened 2026-09-03
Cycle TID values for duplicate records

LeeSaferite:
Doing some data analysis I identified a large number of records using duplicate TID values. These seem to be caused by something like copy/paste of the files or file data.

This group of duplicates were tested to, on the dupe group level, all have the same TID, all use the same Source, have zero Sources referencing them, and all be identical. The repair help minted new TIDs for all dupe group members using the toolbox TID code so they are identical to what GCS would create.

### DIFF
--- a/Library/Aliens/Sparrials/Animals/Chiora - Tamed.gcs
+++ b/Library/Aliens/Sparrials/Animals/Chiora - Tamed.gcs
@@ -583,7 +583,7 @@
 	],
 	"traits": [
 		{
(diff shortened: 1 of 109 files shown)


LeeSaferite:
If it helps in review, `.gcs` and `.gct` files can never be a *source* for another record. This PR contains all `.gcs` and `.gct` files.

richardwilkes (approved):

PR #618, opened 2026-09-03
Removing invailid source references

LeeSaferite:


Library/Aliens/Sparrials/Templates/Hotshot Pilot.gct
```diff
@@ -1576,11 +1576,6 @@
 			"children": [
 				{
 					"id": "T1StNJuJs8Nq-ALQh",
-					"source": {
-						"library": "/gcs_user_library",
-						"path": "Aliens-Sparrials/Aliens-Sparrials Traits.adq",
```
richardwilkes, line 1581: Shouldn't this (and probably the one above, too) be pointing at "Aliens/Sparrials/Sparrials Traits.adq"?

richardwilkes (approved):

